PR #1, opened 2023-10-02
Infrastructure/setup

agnessa:
Adds terraform code to provision the GCP resources:
- network with a VPC connector
- bastion host on Google Compute Engine
- postgres on Cloud SQL
- Artifact Registry repositories
- Cloud Run instance for the cms
- Cloud Run instance for the client
- load balancer
- error reporting
- uptime checks
- GH Secrets

and a GH Actions workflow which:
- builds docker images for the cms and client
- pushes images to Artifact Registry
- deploys the Cloud Run instances

The bastion host is intended for connecting externally to the database for data imports.

About environment secrets:
- the terraform code writes certain secrets, specifically database credentials and the backend url into environment and instance specific GH secrets, whose names are suffixed with "TF_MANAGED", e.g. STAGING_CLIENT_ENV_TF_MANAGED - these are not meant to be changed by humans. They are copied into the respective .env files when building the docker image. It is possible to have human-managed secrets in a separate STAGING_CLIENT_ENV secret - that will be appended when copying the .env file. This allows the client application to have additional configuration, which has nothing to do with the provisioned resources, without going all the way through terraform.


agnessa:
@clementprdhomme can you take a look if you're OK with the FE CD; I added a production Dockerfile, but in fact the only change so far is about the .env file. Please note the strapi instance was just copied from scaffold, nothing interesting going on there yet in terms of endpoints.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,260 @@
...
+#
+#    Cloud Storage
+#      roles/storage.admin   (if using Google Container Registry (gcr) instead)
+#
+#    Artifact Registry
+#      roles/artifactregistry.admin     (project or repository level)
+#
+#    NOTE: You should always follow the principle of least privilege when assigning IAM roles
+#
+# 4. Create GitHub secrets for WIF_PROVIDER and WIF_SERVICE_ACCOUNT
```
clementprdhomme, line 34: I don't understand much of the description above but I'm not seeing these secrets on GitHub. Did you forget to add them?

frontend/Dockerfile.prod
```diff
@@ -0,0 +1,56 @@
+# Install dependencies only when needed
+FROM node:18.15-bullseye AS deps
+
+WORKDIR /app
```
clementprdhomme, line 4: I'm not very familiar with Docker. I just want to double check with you that this folder is in the virtual machine and not the repository because we have no such folder there inside `frontend`.

clementprdhomme (commented):
- Why do you keep track of the environment in the GH Action if we'll only have one environment instead of staging and production?
- Can you detail what's the process to add an extra environment variable for the front-end application? I'm not sure I got it entirely.
- In the longer term, maybe it'd be better for SkyTruth to have the CMS accessible as a folder or something (e.g. maindomain.com/cms).

agnessa:
> * Why do you keep track of the environment in the GH Action if we'll only have one environment instead of staging and production?

If we don't need a second environment, then this won't matter and the cost of having this naming convention is nil, but if it turns out that we do need it, it will help a lot that we have all resources and secrets namespaced already. Also, I followed the way workflows were created for the cms scaffold in AWS and Digital Ocean. I am going to backport this GCP version there, so it makes sense to also keep it consistent with how we usually do things.

> * Can you detail what's the process to add an extra environment variable for the front-end application? I'm not sure I got it entirely.
You can use the `STAGING_CLIENT_ENV` secret to provide all additional env vars, each on a new line like in an `.env` file. I would suggest keeping a source of truth in lastpass. This is appended by the workflow to the env vars managed by terraform. Those are currently these: https://github.com/Vizzuality/skytruth-30x30/blob/a11c26d61f69f50025f2be9e790264d80c9d3132/infrastructure/base/modules/env/main.tf#L160 - happy to just keep the ones related to infrastructure in there, let me know if I should remove something from there so you can redefine in the other env secret.

> * In the longer term, maybe it'd be better for SkyTruth to have the CMS accessible as a folder or something (e.g. maindomain.com/cms).
OK, Tiago had the same comment, I'll try to set it up like that.

frontend/Dockerfile.prod
agnessa: that's fine, yes it's not expected to be in the repository. It is used to set the context for building and will be created if not already there.

.github/workflows/deploy.yml
agnessa: No, they refer to an alternative method of authentication, which we are not using (the workflow comes from a template that showed various options). I'll remove unnecessary comments.

clementprdhomme: No worries. Sorry, I didn't see it was commented 🤦.

PR #8, opened 2023-10-18
[SKY30-82]: adds Orval to client

andresgnlez:
This PR adds [Orval](https://orval.dev/) to the client so we can use generated types from Strapi entities without having to define ourselves.

@agnessa I am adding you to the PR because I am syncing some changes in the CMS I am not 100% sure (I haven't touched anything related with the configuration of the CMS; but still, running the `config-sync` will output some files. Let me know if they are fine or I'll just remove them from the PR.

Jira: https://vizzuality.atlassian.net/browse/SKY30-82


agnessa:
sorry about that, I think it's best if you remove them, because I thought I already synced some of them but clearly that is not the case, so I'd like to check that sync again and see what else I missed and fix in a separate PR

SARodrigues (approved):
LGTM! 

I've tested the new generated types/hooks and they work beautifully.

PR #14, opened 2023-10-19
[SKY30-58]: EZZ layer

andresgnlez:
### Overview
This PR involves prep work to fetch/handle layers available from the CMS and render them in the map.

### Designs

### Testing instructions

### Feature relevant tickets

https://vizzuality.atlassian.net/browse/SKY30-58

---


cms/config/sync/user-role.public.json
```diff
@@ -4,89 +4,17 @@
   "type": "public",
   "permissions": [
     {
-      "action": "api::data-info.data-info.find"
```
agnessa, line 7: This change removed permissions for the FE to most of the data endpoints - please be careful when merging changes to content types, ideally separate them from other changes to make it easier to spot such problems.

PR #30, opened 2023-10-25
Add last updated at value to widgets

SARodrigues:
### Overview

**In this PR:**  
- We display, for each widget, when the data was last updated  
  - Last updated value comes from the API call's metadata  
  - It is formatted to be human readable in a "<Month> <Year>" format, as per designs  

**Notes:**
- I've used `d3-time-format` for parsing the dates. Reasoning: 
  - I've pondered between: 
    - `momentjs`
      _which I particularly don't like, not for any rational reason. But dependencies, etc_
    - `date-fns`
      _we've been using it in other projects_  
    - `d3-time-format`  
 - The only reason I went with `d3-time-format` was because:
   - Smaller size when compared to `date-fns`  
   - We are already using `d3-format` for numbers
     _keeps things consistent_  

**Other notes:**
I took the chance to, for the habitat and marine conservation widgets, change how the data is fetches to make it closer to how the layers are being fetched, both for consistency purposes and I also think it's more readable. 

### Screenshots

Screenshot 2023-10-26 at 14 14 02

### Testing instructions

N/A

### Feature relevant tickets

[SKY30-111](https://vizzuality.atlassian.net/browse/SKY30-111)

[SKY30-111]: https://vizzuality.atlassian.net/browse/SKY30-111?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


andresgnlez:
@SARodrigues can you fix the type issue, please?

frontend/package.json
```diff
@@ -45,6 +45,7 @@
     "clsx": "^2.0.0",
     "cmdk": "^0.2.0",
     "d3-format": "^3.1.0",
+    "d3-time-format": "^4.1.0",
```
andresgnlez, line 48: Fix version, please

SARodrigues: @andresgnlez Can you clarify?

SARodrigues:
@andresgnlez Fixed type issues in the way we found in the call. 
Wasn't as straightforward as expected but it is consistent.

frontend/package.json
andresgnlez: No caveat. Running `yarn add -E d3-time-format` will remove the caveat and fix the version

PR #31, opened 2023-10-26
Strapi connection

aagm:
This PR adds the needed code to connect with the Strapi Backend.


data/src/strapi.py
```diff
@@ -0,0 +1,93 @@
...
+        self.session.headers.update({"Authorization": f"Bearer {jwt}"})
+
+        return self
+
+    def logout(self) -> None:
+        self.session.close()
+
+    @cached_property
+    def getCollections(self):
+        response = self.session.get(f"{self.url}/content-manager/collection-types")
```
agnessa, line 60: sorry, cannot get it to work, maybe you can help :) as far as I can tell, there is no such endpoint `/content-manager/collection-types` (strapi routes:list); there is `/content-manager/collection-types/:model`. There is an endpoint `/content-manager/content-types` which I think would work here, possibly with additional filtering.

data/src/strapi.py
```diff
@@ -0,0 +1,93 @@
...
+        response.raise_for_status()
+        data = [
+            response.json().get("data")
+            for collection in response.json()
```
agnessa, line 64: perhaps you meant to call `.data()` on this line

data/src/strapi.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def getCollectionMetadata(self, collectionName):
+        return list(filter(lambda x: (collectionName in x), self.getCollections))
+
+    def getCollectionData(self, collectionName):
+        response = self.session.get(f"{self.url}/api/{collectionName}")
+        response.raise_for_status()
+        return response.json()
+
+    def importCollection(self, collectionName: str, file_path: Path, idField: str):
```
agnessa, line 77: I didn't get to testing this part, do you have a sample file to use?

PR #34, opened 2023-10-27
[SKY30-87] FE - Navigation to the data table + design tweaks

SARodrigues:
### Overview

This PR adds _"Marine Conservation Details"_ panel to the Data Tool. 

**In this PR:**  
- Added a new _"Marine Conservation Details"_ to the sidebar  
  - Some tweaks required to the sidebar given that the button is stuck to the bottom  
- Added the container for the new _"Details"_ panel  
  - It will not unload the map component; instead it is displayed above it  
    _Some tweaks may be needed in the future to `z-index` when the map tooltips are implemented_  
  - Currently the details panel will only show an existing static table
    _Implementation of the actual table will be addressed by [SKY30-88](https://vizzuality.atlassian.net/browse/SKY30-88); an existing table was used to ensure overflows can work correctly_  
- New hook based with `next-usequerystate` to set up a base for the details/content settings  
  - At the moment in only stores whether the details panel is open
- Some tweaks to approximate the DataTool's visuals to the designs
- New ui/button variants
  - including a new variant for the map's buttons

### Screen Recording

https://github.com/Vizzuality/skytruth-30x30/assets/6273795/67cdbf88-e8e9-48d7-ae23-ecc40aac80e6

### Testing instructions

N/A

### Feature relevant tickets

[SKY30-87](https://vizzuality.atlassian.net/browse/SKY30-87) | FE - Navigation to the data table

[SKY30-87]: https://vizzuality.atlassian.net/browse/SKY30-87?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[SKY30-88]: https://vizzuality.atlassian.net/browse/SKY30-88?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


frontend/src/containers/data-tool/content/details/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { Button } from '@/components/ui/button';
+import { useSyncDataToolContentSettings } from '@/containers/data-tool/sync-settings';
+
+import { mockedData } from './mocked-data';
+
+const DataToolDetails: React.FC = () => {
+  const [settings, setSettings] = useSyncDataToolContentSettings();
+
+  const handleOnCloseClick = () => {
+    setSettings({ ...settings, details: false });
```
andresgnlez, line 11: You can  do `setSettings((prevSettings) => ({ ...prevSettings, details: false }));` so you avoid calling `settings` in the component and unnecessary renders (every time `settings` changes, the component will render)

frontend/src/containers/data-tool/content/index.tsx
```diff
@@ -1,8 +1,17 @@
+import { useSyncDataToolContentSettings } from '@/containers/data-tool/sync-settings';
+
+import Details from './details';
 import Map from './map';
-// import Table from './table';
 
 const DataToolContent: React.FC = () => {
-  return <Map />;
+  const [{ details }] = useSyncDataToolContentSettings();
```
andresgnlez, line 7: A bit picky here but I think renaming the variable to `showDetails` reflects better its nature (you can tell by the name is a boolean), up to you!

frontend/src/containers/data-tool/content/details/mocked-data.ts
```diff
@@ -0,0 +1,67 @@
+import { DashboardTableItem } from '@/components/dashboard-table';
+
+export const mockedData: DashboardTableItem[] = [
```
andresgnlez, line 3: Do we still need mocked data declared on our end? If so, I would create the data in the CMS and pull it as it was real, so we have the hooks and implementation ready and it is a matter of updating the CMS with real data.

SARodrigues: The tables are being done in subsequent tasks.

I've only only added them because we already had these (and the mocked data; it was just copied) already available in the code, so that:  
- The details pane is not empty  
- To test the layout overflows  
The table/data in it doesn't reflect what will be implemented. 

The next task in the sprint (according to the sprint planning breakdown) is [SKY30-88](https://vizzuality.atlassian.net/browse/SKY30-88), which although still uses mocked data, it'll change all this. The code for that particular task is nearly finished, and is based on this PR (because it introduces different tables depending on the location type).

After that we have tasks to start pulling the data from the API and actually populate the tables; I believe by now the endpoints are already available or about to be. :)

frontend/src/containers/data-tool/content/details/index.tsx
SARodrigues: Done. I followed the pattern you used for the map.

frontend/src/containers/data-tool/content/index.tsx
SARodrigues: I agree, `showDetails` is more readable. Done.

SARodrigues:
Pinging @andresgnlez

PR #52, opened 2023-11-13
[SKY30-91] Allow Filtering of the Global/Regional Data Tool table

SARodrigues:
### Overview

**In this PR:**  
- Add base Checkbox ui component  
- Add a reusable "filters" button for use in the data-tool tables   
- Allow global/regional table columns to be filterable (as per designs)    
- Fixed an issue with the column headers Sorting/Filtering buttons being misaligned    
- Fixed an issue with the table displaying a `0` (zero) when no data was available
 
**Notes:**
- The available location types are currently hardcoded in a constants file.
  _the reasoning is because the endpoint for this doesn't yet exist and because we still have to implement another table with more similar options, it stood to reason to just hardcode it for now and then be sure of the number of APIs missing, if necessary. This also allows for backend changes in the meanwhile without breaking the app_  
  - Columns are built into a hook, so pulling data should be easier in the future
- Filtering is done on the frontend  
  _Because the data we display in the table is far from the data we receive the API (we do have to calculate a little), it was just WAY easier to implement the filtering in the frontend. The caveat here is that pagination is off the table for now, but still was with the frontend sorting. Either way, it's not part of the scope._  
  - Current solution is built with an eye for a change to backend filtering in the future
- Typing & components locations could be better: 
  _hopefully improved when implementing the other table as it'll make it more clear what can be shared/simplified_ 

### Feature relevant tickets

[SKY30-91 | FE - Allow filtering the table](https://vizzuality.atlassian.net/browse/SKY30-91)


frontend/src/containers/data-tool/content/details/tables/global-regional/index.tsx
```diff
@@ -13,6 +13,17 @@ const DATA_YEAR = 2023;
 const GlobalRegionalTable: React.FC = () => {
   const location = useAtomValue(locationAtom);
 
+  const [filters, setFilters] = useState({
+    locationType: ['country', 'worldwide', 'highseas', 'region'],
```
SARodrigues, line 17: This was used for testing purposes and I need to remove it.

frontend/src/containers/data-tool/content/details/tables/global-regional/index.tsx
```diff
@@ -13,6 +13,17 @@ const DATA_YEAR = 2023;
...
+  });
+
+  const handleOnFiltersChange = (field, values) => {
+    setFilters({ ...filters, [field]: values });
+  };
+
+  const columns = useColumns({ filters, onFiltersChange: handleOnFiltersChange });
+  // console.log(columns);
```
andresgnlez, line 25: Please remove.

PR #68, opened 2023-11-21
fixes issue loading locations due to pagination

andresgnlez:
### Overview

- This is something to discuss with @agnessa: regardless we are trying to override the limit of pagination in some endpoints, setting `pagination[size]: -1` will, actually, NOT disable the pagination but disable the default one (25) but it will never override the _hard_ limit [set in the API controllers](https://github.com/Vizzuality/skytruth-30x30/blob/develop/cms/config/api.ts#L4). I don't think there's a way to override that default only for a single or specific endpoints, so we might need to come up with something like a custom endpoint to skip this pagination limit.

- Removed the locationAtom: it was being used to store cached data.
- Fixed logic to get location bounds (main reason of the crashes)

### Designs

### Testing instructions

### Feature relevant tickets

-
---


agnessa:
If the problem can be solved by increasing the hard limit, we can just do that - however, not to some crazy number, because next thing that happens is requests are going to time out and instances become unavailable. Otherwise, we can use Tiago's approach from orcasa: https://github.com/Orcasa-Platform/orcasa/commit/bd6f6f32943310ccd3f2f940be1fc49886563a18

andresgnlez:
I think increasing the limit to ~300 should be fine. This is mainly up to the location list: we need that list fully available at once no matter what. Right now, there are around 17X locations. I don't think this number will change too much in the future, hopefully.

So yeah, to keep things I would advocate to increase the general cap to ~300. I am not sure if this can suppose a problem to the stability of the platform, in general we are trying to skip pagination for the tabular data and locations, AFAIK.

SARodrigues (approved):
I see what you meant before about not using the atom for locations as well 👍

PR #74, opened 2023-11-23
[SKY30-150] Small editions to about page

SARodrigues:
### Overview

**In this PR:** 
- In the about page logos grid at the bottom (Team / Funders)
  - We link logos to the respective websites  
  - We simplify the descriptions below the logos

### Feature relevant tickets

[SKY30-150 | Small editions to about page](https://vizzuality.atlassian.net/browse/SKY30-150)


frontend/src/containers/about/logo/index.tsx
```diff
@@ -10,9 +10,9 @@ const Logo: React.FC<LogoProps> = ({ logo: logoKey }) => {
   const { logo, alt, dimensions } = LOGOS[logoKey];
 
   return (
-    <div>
+    <span>
```
andresgnlez, line 13: Why not fragment?

frontend/src/containers/about/logos-grid/types.ts
```diff
@@ -1,7 +1,8 @@
 export type LogoType = {
   logo: string;
   alt: string;
-  description: string;
+  link?: string;
+  description?: string;
   dimensions: [number, number];
```
andresgnlez, line 6: Instead of coming up with your own types, I do recommend always referring to the types of the element/component you are using. For example: we both agree `alt` is a string and the type of `alt` in `Image`, why not just using its typing instead of coming up with your own (and the later maintenance).

Most of components and elements we use already have their own typing, we just need to follow it. The only problem I can find with your approach if, if the future `alt` changes to `string[]`, you will not notice ad your type is fine (it still says `string`), but the actually the image is waiting for `string[]`. If you would have pointed to its type, you would have noticed in the code as it would fail :)

(Ping me if the explanation is not clear or whatever)

image

This is just an example of extracting the type. There are other valid ways.

frontend/src/containers/about/logo/index.tsx
SARodrigues: In most if not all cases in the About page, the parent component will be `flex`. With a fragment (or returning `Image` directly) will cause the logo to be stretched to fill the parent container.

PR #83, opened 2023-11-27
[SKY30-152]: minor visual improvements

andresgnlez:
### Overview

_Check task description to see changes_

### Designs

### Testing instructions

### Feature relevant tickets

https://vizzuality.atlassian.net/browse/SKY30-152

---


SARodrigues (approved):
Only thing I find is the Contact us image. The bottom is cut off; from the designs I would infer that it gets cut off at the bottom if it doesn't fit, but if it does fit then it should be fully visible?

andresgnlez:
The form will be always taller than the image, so it will never cut off with the bottom _naturally_. I would leave it as if.

PR #98, opened 2023-11-29
[N/A] Fix ABNJ data table crash

SARodrigues:
### Overview

Location code used to be `highSeas`, but got changed to `highseas`. 
Data table would fail to get the title.


andresgnlez:
@SARodrigues where we are setting the title, can we fallback in case the key doesn't match so we display a `-` or something similar instead of crashing?

PR #100, opened 2023-11-29
[N/A]: fixes loop with filters in tables

andresgnlez:
### Overview

Changing filters in a `useEffect` is never a good idea 😬 . I've changed it so now the filters update is done in a more imperative way (when the new values is selected). Let me know if I haven't broken/removed anything else 😅 

### Designs

### Testing instructions

### Feature relevant tickets

N/A

---

### DIFF
--- a/frontend/src/containers/data-tool/content/details/table/filters-button/index.tsx
+++ b/frontend/src/containers/data-tool/content/details/table/filters-button/index.tsx
@@ -1,8 +1,7 @@
-import { useEffect, useMemo, useState } from 'react';
+import { useMemo, useState } from 'react';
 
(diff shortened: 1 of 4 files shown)


SARodrigues:
@andresgnlez Looks obvious now.  
It should have been obvious as soon as I saw the issue with dependencies, but it never crossed my mind. 😓 

Small note: the `global-regional` data table has the same issue. Might be worth doing it in the same PR in case we need to revert, but I have found nothing broken; filters work fine!

andresgnlez:
@SARodrigues I almost forget about it, thanks for noticing. I rebased the PR, so the fix for both tables should be in a single commit. Feel free to merge if everything is OK.

PR #107, opened 2023-12-01
[SKY30-163] Create analysis sidebar + widget

SARodrigues:
### Overview

**In this PR:**  
- Slight change to the Sidebar file structure
  _to make space for an "Analysis" sidebar_  
- Add options/props to the HorizontalBarChart chart
  _to make it reusable for the Analysis widget_  
- Reusable tooltip button  
  _still need to change other components to use it, namely the charts_
- Small tweak to the mocked analysis data to make it easier on the FE  
- Analysis Sidebar
  - ~~Intro placeholder~~  Implemented
- Analysis Widget  
  - Widget itself, with charts
    _values from the store_  
    - Includes loading message
    - Includes error message
  - Placeholder for the intro design  
    _SKY30-160_

**Notes:**  
Some tweaks need to be made, although it'll probably be easier when interfacing with the data:  
- Add the intro designs (not yet ready)  
- Adjust the buttons' actions (clear, redraw) - needs to be checked with the team  
- "Loading data" does not appear on a first go; unsure if it's due to how it's currently mocked
- Redrawing on the map does not seem to reset correctly  
- Tooltips  
- Calculations may need to be revised depending on the data we'll have  
- Possibly linking countries to the regular map/details - need to check with the team

### Screenshot  

Screenshot 2023-12-01 at 05 22 39

### Feature relevant tickets

[SKY30-163](https://vizzuality.atlassian.net/browse/SKY30-163)

[SKY30-163]: https://vizzuality.atlassian.net/browse/SKY30-163?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


frontend/src/containers/map/content/map/analysis/index.tsx
```diff
@@ -14,10 +14,11 @@ const fetchAnalysis = async () => {
...
       resolve({
-        data: {
-          locations_area: [{ FRA: 23856, USA: 502367 }],
-          total_area: 736000,
-        },
+        locations_area: [
+          { code: 'FRA', protected_area: 2385406 },
+          { code: 'USA', protected_area: 5000367 },
+        ],
+        total_area: 73600000,
```
andresgnlez, line 21: Have you talked to Alicia about this change?

frontend/src/containers/map/sidebar/analysis/analysis-buttons/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+const AnalysisButtons: React.FC = () => {
+  const setDrawState = useSetAtom(drawStateAtom);
+  const resetAnalysis = useResetAtom(analysisAtom);
+
+  const resetDrawing = useCallback(() => {
+    setDrawState((prevState) => ({
+      ...prevState,
+      feature: null,
+    }));
+  }, [setDrawState]);
```
andresgnlez, line 18: Why are we resetting the atom here manually (and partially) instead of using the `useResetAtom` utility?

.DS_Store
andresgnlez, line 1: Shame on you 😝

frontend/src/containers/map/sidebar/analysis/analysis-buttons/index.tsx
SARodrigues: The flow has changes a bit as per the meet on Tuesday (so these changes are now outdated), and it was uncertain at the time of mock implementation. 

But the idea here was that, when the user clicks _"Re-Draw"_, we would need to clear the feature but still keep the drawing mode active so that the user keeps drawing. 

In the current flow/implementation though, when we reset we now reset the drawing atom, as well as the analysis one, but we still need to keep analysis active. A new property was added to the analysis atom (`active`) for this. Still, the new implementation will first reset the atom and then set `active` back to true; although it works I don't think this is correct, needs another pair of eyes on it if you have the time.

frontend/src/containers/map/content/map/analysis/index.tsx
SARodrigues: Unfortunately by the time I got to this Alicia was off. The endpoint wasn't ready (nor mocked) at this time so I took a couple liberties in order to keep the implementation going; some talk about this [here](https://vizzuality.slack.com/archives/C056QHB3TML/p1701780715046399). 

@agnessa now implemented a quick mocked response in the endpoint matching this; allowing the connection to the endpoint to be made. If we decide to go with the original format it won't be a big deal; a bit more complex on the FE but we can iterate by Object key values still.

SARodrigues:
Rebased on `develop`, added the initial connection to the Analysis endpoint which now mocks a response to get things moving forward. 

Updated Analysis intro implemented (incl. new buttons), tooltips, fixed issues mentioned in the notes.

PR #133, opened 2024-01-16
[SKY30-180][SKY30-183] Use a single decimal place for indicators + display <0.1%

SARodrigues:
### Overview

**In this PR:**  
- Indicators now use a single decimal place  
  _This applies to both the percentages in the widgets, as well as the ones in the tables_  
- If a percentage is bigger than 0 but smaller than 0.1, _"<0.1%"_ will be displayed  

**Notes:**  
I've added the new logic (<0.1%) to the `formatPercentage` helper/util, and changed the widgets' and tables' cell formatter to use the new utility, in order to keep changes to a minimum and the display consistent. 

I've encountered a couple "issues" during this:

On one hand, it looks like the original `formatPercentage` returns the percentage with a percentage sign, which isn't compatible with the designs (as it will not allow for a different styling/size of the percentage sign); in fact I've noticed that right now in production widgets have the percentage sign duplicated. I've added a property to allow the helper to be able to format the numbers both with and without the percentage sign. I'm not 100% confident in this implementation, perhaps we should just use `style: 'decimal'` to keep things simple, though the implementation seems alright. 

On the other hand, something that confused me was that calling `formatPercentage` with, say, `0.0212` in order to display `2.1%` looked counter intuitive. I've changed it so that it'll take `2.12` and display the same `2.1%`, in order to keep things consistent and avoid changing calculations (particularly on the data tables). 

### Feature relevant tickets

[SKY30-180](https://vizzuality.atlassian.net/browse/SKY30-180)
[SKY30-183](https://vizzuality.atlassian.net/browse/SKY30-183)

[SKY30-180]: https://vizzuality.atlassian.net/browse/SKY30-180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SKY30-183]: https://vizzuality.atlassian.net/browse/SKY30-183?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/frontend/src/containers/map/content/map/popup/protected-area/index.tsx
+++ b/frontend/src/containers/map/content/map/popup/protected-area/index.tsx
@@ -133,6 +133,9 @@ const ProtectedAreaPopup = ({ locationId }: { locationId: number }) => {
                 {format({
                   value: DATA?.REP_M_AREA,
                   id: 'formatKM',
+                  options: {
+                    maximumSignificantDigits: 3,
+                  },
                 })}
                 Km<sup>2</sup>
               </dd>
--- a/frontend/src/containers/map/sidebar/details/widgets/marine-conservation/index.tsx
+++ b/frontend/src/containers/map/sidebar/details/widgets/marine-conservation/index.tsx
@@ -4,6 +4,7 @@ import { groupBy } from 'lodash-es';
 
 import ConservationChart from '@/components/charts/conservation-chart';
 import Widget from '@/components/widget';
+import { formatKM } from '@/lib/utils/formats';
 import { useGetProtectionCoverageStats } from '@/types/generated/protection-coverage-stat';
 import type { LocationGroupsDataItemAttributes } from '@/types/generated/strapi.schemas';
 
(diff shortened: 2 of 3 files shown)


andresgnlez (changes requested):
@SARodrigues 
image

Can we keep consistent the styling of the percentage? In the picture, the left one (in the widgets) has more space between the figure and the symbol compared with the one in the table.

The rest looks good to me, thanks!

SARodrigues:
> @SARodrigues image
> 
> Can we keep consistent the styling of the percentage? In the picture, the left one (in the widgets) has more space between the figure and the symbol compared with the one in the table.
> 
> The rest looks good to me, thanks!

Percentage styling was supposed to match the designs. We can keep it consistent, easiest option is to remove the spacing from the percentages in the sidebar. We may have to revisit the designs in the future, which shouldn't be a big deal.

SARodrigues:
@andresgnlez I've merged `develop` into this in order to solve conflicts, also added an extra commit to standardize the spacing between percentage numbers and the percentage sign as suggested.

PR #152, opened 2024-01-24
[SKY30-201] Fix details table headers on scroll

SARodrigues:
### Overview

This PR fixes the details tables' header to the top on scroll. 

**Note:** 
It looks like `position: sticky` doesn't play very well when parent containers have an overflow set, so there was some trickery involved in order to make this work. The headers' borders wouldn't display on scroll either, hence the extra global. 

### Feature relevant tickets

[SKY30-201](https://vizzuality.atlassian.net/browse/SKY30-201)

[SKY30-201]: https://vizzuality.atlassian.net/browse/SKY30-201?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/frontend/src/styles/globals.css
+++ b/frontend/src/styles/globals.css
@@ -37,7 +37,3 @@
 .mapboxgl-popup-anchor-left .mapboxgl-popup-tip {
   @apply !border-r-black
 }
-
-.details-table-first-column-border {
-  border-right: 1px dashed;
-}
--- a/frontend/src/styles/globals.css
+++ b/frontend/src/styles/globals.css
@@ -38,3 +38,6 @@
   @apply !border-r-black
 }
 
+.details-table-first-column-border {
+  border-right: 1px dashed;
+}
--- a/frontend/src/containers/map/content/details/index.tsx
+++ b/frontend/src/containers/map/content/details/index.tsx
@@ -62,8 +62,8 @@ const MapDetails: React.FC = () => {
   }, [locationsQuery.data]);
 
   return (
-    <div className="absolute h-full w-full overflow-x-hidden overflow-y-scroll bg-white px-4 py-4 md:px-6">
-      <div className="mb-8 flex gap-8 md:justify-between">
+    <div className="absolute h-full w-full overflow-scroll bg-white px-4 py-4 md:px-6">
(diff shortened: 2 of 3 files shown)


frontend/src/styles/globals.css
```diff
@@ -38,3 +38,6 @@
   @apply !border-r-black
 }
 
+.details-table-first-column-border {
```
andresgnlez, line 41: Why are you declaring a class here?

frontend/tailwind.config.js
```diff
@@ -49,6 +49,10 @@ module.exports = {
         'accordion-down': 'accordion-down 0.2s ease-out',
         'accordion-up': 'accordion-up 0.2s ease-out',
       },
+      borders: {
+        test: 'border-right-style: dashed',
+        r: { test: 'border-right-style: dashed' },
+      },
```
andresgnlez, line 55: Are you using these?

andresgnlez (changes requested):

frontend/tailwind.config.js
SARodrigues: Uff it looks like I have pushed my test branch, not the correct one. I'll take a look.

frontend/src/styles/globals.css
SARodrigues: Uff it looks like I have pushed my test branch, not the correct one. I'll take a look.

frontend/tailwind.config.js
SARodrigues: Test config removed.

frontend/src/styles/globals.css
SARodrigues: @andresgnlez 

In these first column cells, the bottom border is solid (much like in all cells), but in these particular ones the right border needs to be dashed. I haven't found a way to pull this off the regular tailwind way, hence the custom style. 

Let me know if you know of a way.

andresgnlez:
@SARodrigues let me know if my last commit works for you.

SARodrigues:
@andresgnlez LGTM, thank you. The margin on the right side of the table is a bit off but we can address it afterwards

PR #157, opened 2024-02-02
[SKY30-223] Make links within info buttons/tooltips clickable

SARodrigues:
### Overview

This PR makes it so that links within info buttons/tooltips are clickable. 

This applies to:  
  - Widgets / charts tooltips  
  - Table column headers  

This is achieved by the `react-linkify` library. In addition, the details tables' column info buttons now make use of the global tooltip component, so that any related changes to this are applied app-wide. 

### Testing instructions

1. Visit the map page
2. On the _"Marine Conservation Coverage"_
    - Verify the _"30x30 Target"_ Info button's link is clickable
3. On the _"Marine Conservation Protection Levels"_ bar charts
    - Verify the info buttons' links are clickable
4. Open the details' table
    - Verify that links are clickable in the table column headers' info buttons 

### Feature relevant tickets

[SKY30-223](https://vizzuality.atlassian.net/browse/SKY30-223)

[SKY30-223]: https://vizzuality.atlassian.net/browse/SKY30-223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -65,6 +65,7 @@
     "react-dom": "18.2.0",
     "react-hook-form": "^7.48.2",
     "react-icons": "4.11.0",
+    "react-linkify": "^1.0.0-alpha",
     "react-map-gl": "7.1.6",
     "recharts": "^2.9.0",
     "rooks": "7.14.1",
--- a/frontend/src/components/tooltip-button/index.tsx
+++ b/frontend/src/components/tooltip-button/index.tsx
@@ -1,5 +1,7 @@
 import { useState } from 'react';
 
+import Linkify from 'react-linkify';
+
 import { Info } from 'lucide-react';
 
 import { Button } from '@/components/ui/button';
@@ -28,7 +30,23 @@ const TooltipButton: React.FC<TooltipButtonProps> = ({ className, text }) => {
         align="center"
         className="flex max-w-[300px] flex-col gap-6 font-mono text-xs"
       >
-        {text}
+        <span>
+          <Linkify
+            componentDecorator={(href, text, key) => (
+              <a
+                className="break-all"
+                href={href}
+                key={key}
+                target="_blank"
+                rel="noopener noreferrer"
+              >
+                {text}
+              </a>
+            )}
+          >
+            {text}
+          </Linkify>
+        </span>
       </PopoverContent>
     </Popover>
   );
--- a/frontend/src/components/tooltip-button/index.tsx
+++ b/frontend/src/components/tooltip-button/index.tsx
@@ -34,7 +34,7 @@ const TooltipButton: React.FC<TooltipButtonProps> = ({ className, text }) => {
           <Linkify
             componentDecorator={(href, text, key) => (
               <a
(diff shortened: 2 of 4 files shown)


andresgnlez:
@SARodrigues the implementation looks good. I am worried about the fact the library is in _alpha_ and has not been updated in 5 years.. The [original library](https://github.com/markdown-it/linkify-it) is maintained though.

Also, we are missing text links (just displaying raw links) and no styling to make the links more prominent. Can you at least add underline styling (or whatever we have in design) to the links, please?

andresgnlez (changes requested):
@SARodrigues the implementation looks good. I am worried about the fact the library is in alpha and has not been updated in 5 years.. The [original library](https://github.com/markdown-it/linkify-it) is maintained though.

Also, we are missing text links (just displaying raw links) and no styling to make the links more prominent. Can you at least add underline styling (or whatever we have in design) to the links, please?

SARodrigues:
> @SARodrigues the implementation looks good. I am worried about the fact the library is in alpha and has not been updated in 5 years.. The [original library](https://github.com/markdown-it/linkify-it) is maintained though.

That caught my eye as well when looking for easy to use libraries, but I'm not sure whether going with `linkify-it` would be worth it here, given the use case and for time purposes. 

Although this one isn't maintained, it has over 200k weekly downloads and fulfills our use case quite well and is extremely simple to use. 

Going with the original would (IMHO) mean spending the time to create a custom component that re-implements functionality that already exists. Right now we'd have at very least to duplicate part of the [string parsing](https://github.com/tasti/react-linkify/blob/master/src/components/Linkify.jsx#L26-L58) functionality in order to support multiple links in the same string coming from Strapi, and in the event we happen to need to support have a need to support multiple child elements we'd need to duplicate [even more functionality](https://github.com/tasti/react-linkify/blob/master/src/components/Linkify.jsx#L60-L70).

What do you think? 

&nbsp;

> Also, we are missing text links (just displaying raw links) and no styling to make the links more prominent. Can you at least add underline styling (or whatever we have in design) to the links, please?

Added an underline to the links; the designs are a bit scarce on this one but I agree we should have it for consistency and user friendliness.

andresgnlez:
> Going with the original would (IMHO) mean spending the time to create a custom component that re-implements functionality that already exists. Right now we'd have at very least to duplicate part of the [string parsing](https://github.com/tasti/react-linkify/blob/master/src/components/Linkify.jsx#L26-L58) functionality in order to support multiple links in the same string coming from Strapi, and in the event we happen to need to support have a need to support multiple child elements we'd need to duplicate [even more functionality](https://github.com/tasti/react-linkify/blob/master/src/components/Linkify.jsx#L60-L70).

> What do you think?

Yeah, I totally agree, it's not worth it. Let's stick with the wrapper. Worst case scenario is the wrapper is not too spread in the codebase so changing it shouldn't be painful.

PR #244, opened 2024-04-10
[SKY30-280] Protection Levels widget area changes

SARodrigues:
### Overview

This PR applies to both the Protection Levels widget, and the Fishing Protection one (recently split to a separate widget). 

This PR changes the areas display (and this calculations), to calculate areas based on the area within the classification (mpaa, fishing), and display the total area within the classification. 

**Note:** I think this could have been implemented better, but given the changes to the query (necessary to get all areas within the classification so we can calculate the total), as well as the calculation changes, I tried to deviate as little as possible from the original code. Both are calculated similarly, so I kept them similar still, in case we need to do further tweaks. 

**Note:** As I understand we will be potentially hiding the Fishing Protection widget soon, but I've updated the calculations due to the similarity, and to prevent confusion in the future. 

### Screenshot  

Screenshot 2024-04-10 at 14 23 04

### Feature relevant tickets

[SKY30-280](https://vizzuality.atlassian.net/browse/SKY30-280)

[SKY30-280]: https://vizzuality.atlassian.net/browse/SKY30-280?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/frontend/src/containers/map/sidebar/main-panel/details/widgets/fishing-protection/index.tsx
+++ b/frontend/src/containers/map/sidebar/main-panel/details/widgets/fishing-protection/index.tsx
@@ -24,11 +24,6 @@ const ProtectionTypesWidget: React.FC<ProtectionTypesWidgetProps> = ({ location 
       // @ts-ignore
       populate: {
         fishing_protection_level_stats: {
-          filters: {
-            fishing_protection_level: {
-              slug: 'highly',
-            },
-          },
           populate: {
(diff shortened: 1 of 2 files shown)


SARodrigues:
We decided that staging values are correct; no changes needed. Keep the PR for future reference

PR #276, opened 2024-06-24
Push and pull translations from Localazy, language selector

clementprdhomme:
This PR does the following:

- Add an action to push the static source strings to Localazy (`latest` release tag) when changes are made in the `develop` branch
- Add an action to promote the translations from the `latest` release tag to the `production` release tag when changes are pushed to the `main` branch
- Fetch the translations from the CDN using the `LOCALAZY_CDN` URL or else use the local files
- Fix redirection and visual issues when viewing the content in Spanish or French
- Add the language selector in the header
- Fix a missing dependency for the translations (in the national high seas table)

Localazy has been configured so there are two releases: _latest_ (default) and _production_. Staging will use the _latest_ translations while production, the _production_ ones. This is configured via the `LOCALAZY_CDN` environment variable that points to the _latest_ or _production_ metadata file on the CDN.

When strings are pushed to localazy, the release is not specified so they will be pushed to the _latest_ release. When making changes in the `main` branch, the GitHub Action moves all the translations from the _latest_ release over to the _production_ release so that the production app can use both the latest code and latest translations.


SARodrigues (approved):

SARodrigues (commented):
@clementprdhomme I was taking a second look at this locally, shouldn't it be `files[1]`?  

I ask because looking at the CDN metafile, `files[0]` is the Strapi one, while `files[1]` seems to be the one containing the translations for the FE. 🤔

.github/workflows/localazy-release.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  localazy-release:
+    name: Release strings to production on Localazy
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: localazy/tag@v1
+        with:
+          workdir: 'frontend'
+          read_key: ${{ secrets.b }}
```
SARodrigues, line 17: @clementprdhomme Should it be `secrets.LOCALAZY_READ_KEY` or was this intentional?

PR #277, opened 2024-07-02
Translate the Strapi content with Localazy

clementprdhomme:
This PR translates all the Strapi content with Localazy. It can't be merged before #276.

Issue: https://vizzuality.atlassian.net/browse/SKY30-375


clementprdhomme:
@SARodrigues I merged `develop` (Alicia's changes) into this PR and they are no other changes to make *but* she removed one model that is in use in the tables so the app is currently not compiling.

Here's the file that causes the issue: `src/containers/map/content/details/tables/national-highseas/index.tsx`

SARodrigues:
@clementprdhomme Just so you are aware and we keep track of the flow of things, there is a PR open to adapt the FE for Alicia's changes on `develop`, https://github.com/Vizzuality/skytruth-30x30/pull/281 . 

It's very detached from translations (I think) and I think would unblock PRs/tasks. We're still working out some stuff in JIRA/GitHub/Slack but after that we should be able to merge your changes. 

Caveat is that the database migrations aren't working as expected (from a FE perspective), but we're sorting a solution.

PR #281, opened 2024-07-11
[SKY30-418] Adapt national highseas tables to work with new mpaa strapi changes

SARodrigues:
### Overview

This PR tweaks the national table to make use of the new query introduced by #278, in order to support the adaptations to the MPA tables adaptation. 

**Notes:**

- The table will _still_ display empty at this point. 
  _Looking at the entries in Strapi, it looks like the MPAs data isn't fully complete; the associations with locations, etc is not yet done._
  _We can test this by visiting the following URL: https://30x30-dev.skytruth.org/cms/api/mpas?filters[location][code][$eq]=GBR, or check it on Strapi_

- There may be some issues testing this locally, as migrations may fail
  _It looks like properties (such as `bbox` or `is_child`) in Strapi are mandatory, but they are not set up._
  _I've also faced issues with `wdpaid` types on Strapi's migrations._
  _The way I worked around this was to wipe my local DB, then export/import each table individually from staging to my local setup. This also involved manipulating some JSON files (most notably for MPAs) to get around the mandatory `bbox` and `is_child`_

- I've also made a small tweak to the tables' display, in order to force them to take _at least_ the full available horizontal space. 
  _This was mostly so that they don't display as weirdly when there is no data._  
  _When #266 is merged, they'd display even more weirdly due to the addition of the lateral scrolling indicators._

&nbsp;

### Screenshot

Screenshot 2024-07-11 at 11 25 56

&nbsp;

### Feature relevant tickets

[SKY30-418](https://vizzuality.atlassian.net/browse/SKY30-418)

[SKY30-418]: https://vizzuality.atlassian.net/browse/SKY30-418?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aagm:
Hi Simao!
Yes data is not yet uploaded, the best thing i can do to reduce the noise that we could have is to delete the entries on the mpa table

Regarding your local, any change in bd unfortunatly needs to be dealt manually, as migrations in strapi, aren't properly managed. My recomendation, if possible is to work against staging if possible, and to use a clean db.

SARodrigues:
@aagm Thank you for the reply! 

I don't think it's worth the time to delete entries in the MPA table. Goal of this PR is just really to use the new endpoint, which when tested locally can work, and unblock the other PRs. Once the data is there things should start working, from my testing.

PR #300, opened 2024-09-06
Add tabs to switch to the Summary and Terrestrial views

clementprdhomme:
This PR adds tabs to the sidebar to allow the user to select between one of 3 views:

- Summary (default)
- Terrestrial
- Marine

At the moment, only the Marine tab has some data (the previous content from the sidebar).

The tabs also affect the table view which right now, as for the sidebar, only displays content when the Marine tab is active.

The active tab is stored in the URL and is remembered when changing the location.

## Testing instructions

- Use your local database as this PR is based on the `terrestrial` branch that has a different database schema
- Comment the `LOCALAZY_CDN` environment variable in your local `.env` to see the source strings (they are not uploaded to the CDN)

## Tracking

[SKY30-385](https://vizzuality.atlassian.net/browse/SKY30-385)


SARodrigues (approved):
Tested locally, LGTM including tab persistence!

PR #301, opened 2024-09-09
Address issue where the data displayed in the global table would be different depending on locale

clementprdhomme:
This PR addresses an issue where the global table would be empty on staging when displayed in English or French, but complete when displayed in Spanish.

The root cause is how Strapi gives a different ID to different version of a same localised document.

The stats models are not localised and have a many-to-1 relation to the Location model, meaning they can only point to one version of a location instead of all them. For this reason, when the stats are loaded through the Location model, only one of the localised version has the link to the stats.

This PR mitigates the issue by loading all three localised version of a Location and determining which one has the relation to the stats model(s). This workaround had already been implemented for the Fishing Protection widget but not replicated for the global table.

## Testing instructions

In the global view, open the table. Switch to all the different locales. Make sure there is always data shown in the table i.e. that the numbers are not all 0.

## Tracking

[SKY30-446](https://vizzuality.atlassian.net/browse/SKY30-446)

[SKY30-446]: https://vizzuality.atlassian.net/browse/SKY30-446?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/frontend/src/containers/map/content/details/tables/global-regional/index.tsx
+++ b/frontend/src/containers/map/content/details/tables/global-regional/index.tsx
@@ -53,7 +53,12 @@ const GlobalRegionalTable: FCWithMessages = () => {
   // Get location data and calculate data to display on the table
   const { data: locationsData }: { data: LocationListResponseDataItem[] } = useGetLocations(
     {
-      locale,
+      // We will use the data from the `localizations` field because the models “Protection Coverage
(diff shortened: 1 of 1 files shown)


SARodrigues (approved):
Tested locally as per instructions, LGTM. Code as well, looks like it was hard to figure out 👍

PR #314, opened 2024-10-08
Layers panel to display datasets grouped by environment - connect to Strapi

SARodrigues:
### Overview

This PR connects the layers panel to the actual data coming from Strapi, and displays layers by environment type based on which environment a layer is currently assigned to. 

Changes were planned to be unobtrusive, and confined to the `LayersPanel` component that fetches datasets; it fetches datasets basically the same way as before, though it does populate the `environment` for the `layers` content type, and then processes them for display. The `LayersGroup` component is then kept unchanged. 

**Logic:**  
- Datasets are pulled in the same way as before, but now `environment` is populated as well  
- Datasets are parsed into 3 groups for display  
  - Basemap (which is environment agnostic, but it is defined as a dataset group in Strapi)  
    - On the display, we are injecting the `Labels` as a toggle, hence the separate processing, same as before
  - Marine + Terrestrial groups (datasets that are not Basemap)  
    - For each environment (Marine, Terrestrial), we parse the dataset, in the following way:  
      - For each dataset, we filter the corresponding layers by the environment  
      - If a dataset does not have any layers (after filtering by environment), the dataset itself is filtered out  

**Result:**  

This allows for correct display, isolation of the dataset/layer parsing for display, and some flexibility:  

- We don't have to create different datasets with the same name for both Maritime and Terrestrial groups;
  - We have an environment agnostic dataset, it is the layers themselves that can belong to an environment  
  - It will allow the FE to display a specific layer for both Terrestrial and Maritime groups in the future  
    - Right now this is not possible due to the relationships established in the BE/Strapi, but the FE would allow this if it were to change  

**Caveats:**  

A layer _must_ have an environment associated on Strapi, or it won't be displayed as a toggle anywhere. We're filtering layers by environment, so if one doesn't have one associated, then the FE cannot display it (though the application will not break. The map, however, will display it. This does mean we could potentially display map layers that do not have corresponding toggles on the panel. 
  

### Testing instructions

1. Test environment / staging (with current data):    
  - Simply change a layer's environment from "Maritime"  to "Terrestrial" (on Strapi)  
    - It should automatically be displayed as a Terrestrial layer, dataset group name should display as well  
2. Locally with (local test data):  
  - Create a new test layer  
  - Create a new "Terrestrial" dataset
    _in order to test a new header entry_  
  - Assign the relationship and dataset  
  - It should be displayed (eg: Screenshot below)

### Screenshot

Screenshot 2024-10-08 at 13 43 57

### Feature relevant tickets

[SKY30-444](https://vizzuality.atlassian.net/browse/SKY30-444)

[SKY30-444]: https://vizzuality.atlassian.net/browse/SKY30-444?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/frontend/src/constants/datasets.ts
+++ b/frontend/src/constants/datasets.ts
@@ -1,0 +1,5 @@
+export const DATASETS = {
+  terrestrial: 'terrestrial',
+  marine: 'marine',
+  basemap: 'basemap',
+};
--- a/frontend/src/constants/environments.ts
+++ b/frontend/src/constants/environments.ts
@@ -1,5 +1,4 @@
-export const DATASETS = {
+export const ENVIRONMENTS = {
   terrestrial: 'terrestrial',
   marine: 'marine',
-  basemap: 'basemap',
 };
--- a/frontend/src/containers/map/sidebar/layers-panel/index.tsx
+++ b/frontend/src/containers/map/sidebar/layers-panel/index.tsx
@@ -174,6 +174,7 @@ const LayersPanel: FCWithMessages = (): JSX.Element => {
         loading={isFetchingDatasetsData}
         showDatasetsNames={false}
         showBottomBorder={false}
+        extraActiveLayers={labels ? 1 : 0}
       >
         {/*
           The labels toggle doesn't come from the basemap dataset and has slightly functionality implemented.
--- a/frontend/src/containers/map/sidebar/layers-panel/index.tsx
+++ b/frontend/src/containers/map/sidebar/layers-panel/index.tsx
@@ -1,10 +1,11 @@
-import { ComponentProps, useCallback } from 'react';
+import { ComponentProps, useCallback, useMemo } from 'react';
 
 import { useLocale, useTranslations } from 'next-intl';
 
 import TooltipButton from '@/components/tooltip-button';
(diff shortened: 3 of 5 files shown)


clementprdhomme (commented):
All your changes are ok but in order to finish the work on the layers panel, we'd need some more work:

- When viewing the “Summary” tab, all the groups are expanded
- When switching tab, reset the list of active layers to the default

That last point is a bit more complex than it seems. When the user is on the “Summary” tab, the user should see the default layers of all 3 groups. When the user is on the “Terrestrial” tab, they should only see the default layers of the Terrestrial, and so on for the “Marine” tab.

That means that when the tab changes, you have to change which layers are active.

In order to know which layers should be active, you can look at the `default` field on the Layer model.

clementprdhomme:
@SARodrigues I added two tiny commits:

1. One to make sure we're using the correct font for the loading and no-data message within a layer group
2. The other to make sure we count the “Labels” layer when the Basemap group is collapsed

SARodrigues:
@clementprdhomme LGTM! 
Feel free to merge!

PR #330, opened 2025-01-30
Tech 2492: Fix conservation builder typo and add to README's

wlytle:
### Overview
This PR fixes a small typo on the Conservation Builder info panel and adds more info to the README's. It also fixes a jitter with the conservation builder info panel when scrolling. This scroll jitter bug was caused by the panel title shrinking on scroll, which - on certain sized screens - made the panel not need to scroll which lurched the title back to a larger size. It was resolved by resizing the title only if the screen size is below a threshold - at this time that is the threshold where the scroll bar would exist. 

Current:
https://github.com/user-attachments/assets/401b5530-5be3-4d4c-9fed-81050fe72c0f

This Change:
https://github.com/user-attachments/assets/f87edfca-e3c5-4c14-9fa6-ffb84fac696c

### Designs

### Testing instructions

### Feature relevant tickets

https://skytruth.atlassian.net/browse/TECH-2492
---


cms/config/sync/admin-role.strapi-super-admin.json
```diff
@@ -16,41 +17,44 @@
         ],
         "locales": [
           "en",
+          "fr",
```
wlytle, line 20: @clementprdhomme I'm not sure what caused these slight changes here? Should I be concerned about this? The duplication of `fr` seems not great.

clementprdhomme: Yep, that should not happen. I wonder if somehow you have two French locales in `/cms/admin/settings/internationalization` locally?

One quick fix you could do is to revert this file and then import the file-based configuration to your locale database in Settings → Config Sync / Interface → Import. Then when you commit again, this file should not have the duplicate locale.

Let me know if that fixes the issue.

aemonm (approved):

PR #336, opened 2025-02-25
TECH-2705: Attempted fix for Google Tag Manager

wlytle:
### Overview

This PR updates how GTM is implemented, since it is currently not logging users to google analytics.

It appears to work locally in debug mode, but GA is set to accept streaming only from the prod site so there may be follow up PR's after testing in Prod

---

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "skytruth-30x30-frontend",
-  "version": "0.0.2",
+  "version": "0.0.3",
   "private": true,
   "scripts": {
     "dev": "yarn types && next dev",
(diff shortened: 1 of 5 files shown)


frontend/package.json
```diff
@@ -29,6 +29,7 @@
     "@loaders.gl/zip": "^3.4.14",
     "@localazy/cdn-client": "^1.5.2",
     "@mapbox/mapbox-gl-draw": "1.4.3",
+    "@next/third-parties": "latest",
```
aemonm, line 32: This reminds me I should probably switch back to using this in Flaring and Cerulean - we went through a debugging phase and I think we landed on adding the scripts directly (but I suspect it will work fine using the recommend package).

aemonm (approved):

frontend/package.json
wlytle: Locally I tried it both ways and it only seems to work with both 🤷 I dunno I'm still very much in the debugging phase of this

PR #343, opened 2025-03-13
TECH-2791: Support for external API use

wlytle:
### Overview

This PR does 2 main things

1. Updates the location to PAs relationship in strapi so that a location has many PA's instead of PA belongs to a location. This allows users to get Data Source information via a query to something like protected area stats: `protected area stats -> location -> PA -> data source` This support users being able to cite the data they get form us.
2. Adds API documentation

### Feature relevant tickets

https://skytruth.atlassian.net/browse/TECH-2791
---

### DIFF
--- a/cms/Documentation.md
+++ b/cms/Documentation.md
@@ -1,0 +1,135 @@
+# 30x30 Progress Tracker REST API Documentation
+
+## Overview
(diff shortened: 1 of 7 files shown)


aemonm:
Cool! would be interested to know how this API works whenever you get your head around it.

wlytle:
> Cool! would be interested to know how this API works whenever you get your head around it.

Finally feel like I have a general idea of what's going. Happy to show you what I've learned

PR #346, opened 2025-03-24
  TECH-2497: Update sorting logic for PA coverage stats

wlytle:
### Overview

This PR updates the way the API sorts  data for Protected Area Coverage Stats. It is primarily targeted at the `fully/highly protected` column, as that is the only on that can theoretically be null and is sorted on (other potential null columns, like `IUCN Category` only get filtered on). This forces `null` values to always sort to the bottom of the data. 

Also there's some documentation clean up

### Testing instructions

Navigate to https://30x30-dev.skytruth.org/progress-tracker  , select the `Marine` tab from the right hand display and then select a Region, or Global as the location, then select `Show Marine Insights Table` and sort on the `Fully/Highly Protected` column. Rows listed as `No Data Available` should always be at the bottom, and the order should be preserved even after selecting the next page of data.

### Feature relevant tickets

https://skytruth.atlassian.net/browse/TECH-2497

---


cms/src/api/protection-coverage-stat/controllers/protection-coverage-stat.ts
```diff
@@ -4,21 +4,116 @@
...
+        try {
+            const { query } = ctx;
+            // find the most recently updated record and return its updatedAt date
+            const updatedAtQuery = {
+                ...query,
+                fields: ['updatedAt'],
+                sort: { updatedAt: 'desc' },
+                limit: 1
+            };
+            const updatedAt = await strapi.entityService.findMany('api::protection-coverage-stat.protection-coverage-stat', updatedAtQuery).then((data) => {
```
aemonm, line 24: Given that is string is used above, and it's meaning is a little opaque, might be cleaner to have it stored as a constant with a simpler name.

cms/src/api/protection-coverage-stat/controllers/protection-coverage-stat.ts
```diff
@@ -4,21 +4,116 @@
...
 }));
+
+interface Pagination {
+    start?: number;
+    limit?: number;
+    page?: number;
+    pageSize?: number;
+}
+
+function getPaginationBounds(pagination: Pagination, dataLength: number): [number, number] {
```
aemonm, line 83: Not necessary here as I think it's easy to understand from context, but a cool TS feature I learned recently (and should definitely be using more of) is [named tuples](https://www.typescriptlang.org/play/?ts=4.0.2#example/named-tuples).

I have a lot of places for things like lat, lon or bounding boxes where it's hard to infer the order without checking the docs, but if I had `[lon: number, lat: number]` or `[minX: number, maxX: number, minY: number, maxY: number]` I could save myself the trip.

cms/src/api/protection-coverage-stat/controllers/protection-coverage-stat.ts
```diff
@@ -4,21 +4,116 @@
...
+                    })
+
+                    sortIndex += 2;
+                }
+            }
+            const [start, end] = getPaginationBounds(query.pagination, data.length);
+            const paginatedData = data.slice(start, end);
+            return { data: paginatedData, meta: { ...meta, updatedAt } };
+        } catch (error) {
+            return ctx.badRequest('Something went wrong', error);
```
aemonm, line 71: I'm not familiar with Strapi, but does the error here get logged? I'm in the beginning stages of trying to improve error handling in Cerulean and my general approach is `console.error` the raw error message and return a meaningful message to the user. 

I'm generally trying to a) make tracking down errors/bugs easier for me and b) not potentially leak sensitive backend details (which might be contained in the raw error object) to users who at best probably don't care. Totally open to other approaches or opinions though.

aemonm (approved):
lgtm

cms/src/api/protection-coverage-stat/controllers/protection-coverage-stat.ts
wlytle: Done

cms/src/api/protection-coverage-stat/controllers/protection-coverage-stat.ts
wlytle: That's rad, thanks for sharing! I added it in.

cms/src/api/protection-coverage-stat/controllers/protection-coverage-stat.ts
wlytle: Good call out. That error was being logged as a `400` response in google, but just the default google log that acknowledges the status, so not much info. I wrote a quick logger class for future me to use and actually logged the error along with removing the error object from the api response.

PR #349, opened 2025-03-28
Hotfix: Handle sort parameter as object

wlytle:
### Overview

This PR is a hot fix for a bug when calling the get protection coverage stats endpoint. It breaks when the sort query param is parsed as an object instead of a string.

### Designs

### Testing instructions

### Feature relevant tickets

---

### DIFF
--- a/cms/src/api/protection-coverage-stat/controllers/protection-coverage-stat.ts
+++ b/cms/src/api/protection-coverage-stat/controllers/protection-coverage-stat.ts
@@ -40,12 +40,23 @@ export default factories.createCoreController(PROTECTION_COVERAGE_STAT_NAMESPACE
             const { data, meta } = await super.find(ctx);
             // Update sort so null values are at the end
             if (query?.sort) {
-                const sortParams = query.sort.split(/[,:]/);
-                let sortIndex = 0;
-                while (sortIndex < sortParams.length) {
+                let sortParams = {}
+                if (typeof query.sort === 'string') {
+                    const sortParamsArray = query.sort.split(/[,:]/);
+
+                    sortParamsArray.forEach((param, index) => {
+                        if (index % 2 === 0) {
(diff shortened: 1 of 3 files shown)


cms/src/api/protection-coverage-stat/controllers/protection-coverage-stat.ts
```diff
@@ -40,12 +40,23 @@ export default factories.createCoreController(PROTECTION_COVERAGE_STAT_NAMESPACE
             const { data, meta } = await super.find(ctx);
             // Update sort so null values are at the end
             if (query?.sort) {
-                const sortParams = query.sort.split(/[,:]/);
-                let sortIndex = 0;
-                while (sortIndex < sortParams.length) {
+                let sortParams = {}
+                if (typeof query.sort === 'string') {
```
aemonm, line 44: Is the issue sometimes the `query.sort` is parsed into an object and sometimes it's left as a string? Or are you dealing with two different `query.sort` entities entirely and this is a central way to deal with both types.

Seems like a lot of hoops to jump through! I'm not at all familiar with the setup or all the cases you need to cover.

aemonm (approved):

cms/src/api/protection-coverage-stat/controllers/protection-coverage-stat.ts
wlytle: I'm also not that familiar with the set up, I thought it all got parsed as a string, hence the error. What appears to be happening is that sorting is being done by the code using the query param structure `sort[year]=desc` and `sort=year:asc` both are apparently cool with strapi but the former parses as `{year: 'desc'}` and the latter parses as `'year:desc'` Ultimately it would be nice to make all the code handle sorting the same way, but fixing this hear I felt was both important and faster than seeking out all the api calls

PR #350, opened 2025-03-28
Update sorting to handle array sort params

wlytle:
### Overview

This PR further updates the handling of sort params for the protection coverage stats controller. Sort params can be passed a  string

```javascript
'location.mpaa_protection_level_stats.percentage:desc,environment.name:asc'
```
An object
```javascript
{
  location: {
    mpaa_protection_level_statspercentage: 'desc'
  },
  environment: {
    name: 'asc'
  }
{
```
or an Array with entries that could be either of the above

```javascript
[
  'location.mpaa_protection_level_stats.percentage:desc',
  {
     environment: {
      name: 'asc'
    }
  }
]

This updates the method to handle all three cases.

```

### Designs

### Testing instructions

### Feature relevant tickets

---

### DIFF
--- a/cms/src/api/protection-coverage-stat/controllers/protection-coverage-stat.ts
+++ b/cms/src/api/protection-coverage-stat/controllers/protection-coverage-stat.ts
@@ -158,7 +158,7 @@ function flattenSortParams(
     const flattened: Record<string, string> = {};
 
     for (const key in sort) {
-        if (Object.prototype.hasOwnProperty.call(sort, key)) {
+        if (Object.hasOwn(sort, key)) {
             const value = sort[key];
             const newKey = prefix ? `${prefix}.${key}` : key;
             if (value !== null && typeof value === 'object') {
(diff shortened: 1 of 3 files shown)


cms/src/api/protection-coverage-stat/controllers/protection-coverage-stat.ts
```diff
@@ -154,3 +142,78 @@ function getValueByPath(
...
+ * @example { 'location.name': 'asc', environment: 'asc' }
+ */
+function flattenSortParams(
+    sort: Record<string, any>,
+    prefix: string = ''
+): Record<string, string> {
+    const flattened: Record<string, string> = {};
+
+    for (const key in sort) {
+        if (Object.prototype.hasOwnProperty.call(sort, key)) {
```
aemonm, line 161: Any reason not to just use `sort?.hasOwnProperty(key)`?

aemonm: Oh, now there's also: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn

aemonm (approved):
lgtm

cms/src/api/protection-coverage-stat/controllers/protection-coverage-stat.ts
wlytle: > Any reason not to just use `sort?.hasOwnProperty(key)`?

Good question. I have this in my head as the "better" way to do it, but I can't recall why. I think something like the object no being crated by this code could have the `hasOwnProperty` method monkey coded as something else... 
I'm really not sure. I didn't know about `hasOwn` which seems to be the latest and greatest iteration, so I've just updated to that. Thanks for the info!

PR #351, opened 2025-04-07
TECH-2494: Update MPAA  protection level filter options

wlytle:
### Overview

The protection level filter menu is populated from all protection levels in the DB. 2 of these levels are composite levels, that are currently not in use but could be used for internal metric aggregation. Either way, these two composite levels, do not correspond to any real MPA and hence don't make sense to offer as filter options. 

This PR, also removes some filtering on these unused composite levels in API calls

### Testing instructions

### Feature relevant tickets

---

### DIFF
--- a/frontend/src/containers/map/content/details/tables/national-highseas/hooks.tsx
+++ b/frontend/src/containers/map/content/details/tables/national-highseas/hooks.tsx
@@ -235,11 +235,12 @@ const useFiltersOptions = () => {
              */
             if (
               mpaaProtectionLevel.attributes.slug !== 'fully-highly-protected' &&
-              mpaaProtectionLevel.attributes.slug !== 'less-protected-unknown') {
+              mpaaProtectionLevel.attributes.slug !== 'less-protected-unknown'
+            ) {
               acc.push({
                 name: mpaaProtectionLevel.attributes.name,
                 value: mpaaProtectionLevel.attributes.slug,
-              })
+              });
             }
             return acc;
           }, []),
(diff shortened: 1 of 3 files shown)


aemonm (approved):

frontend/src/containers/map/content/details/tables/national-highseas/hooks.tsx
```diff
@@ -228,10 +228,21 @@ const useFiltersOptions = () => {
     {
       query: {
         select: ({ data }) =>
-          data.map((mpaaProtectionLevel) => ({
-            name: mpaaProtectionLevel.attributes.name,
-            value: mpaaProtectionLevel.attributes.slug,
-          })),
+          data.reduce((acc, mpaaProtectionLevel) => {
```
aemonm, line 231: tiny personal preference would be to favor the `map().filter()` chain, but just every so tiny preference as I think it's slightly more explicit in what the intent is - but totally great as is!

wlytle: Oh I like this. I went with reduce because then it only iterates through the array once, but this array only has like 8 entries in it ever so a double iteration is trivial. I'll swap it out in a bit.

wlytle: Updated, for sure easier to read now.

wlytle: Oh cool, thanks for the info. I feel like the filter.map chain is most readable for this simple option, but I'll keep `flatMap` in my back pocket

PR #353, opened 2025-04-23
Test tf stack

sstill88:
### Overview

_Please write a description. If the PR is hard to understand, provide a quick explanation of the code._

### Designs

### Testing instructions

### Feature relevant tickets

---

### DIFF
--- a/cloud_functions/test/Dockerfile
+++ b/cloud_functions/test/Dockerfile
@@ -1,9 +1,6 @@
 # Use the official Python image
 FROM python:3.9-slim
 
-# Set environment variables
-ENV PYTHONUNBUFFERED=1
-
 # Set working directory
 WORKDIR /app
 
--- a/cloud_functions/test/Dockerfile
+++ b/cloud_functions/test/Dockerfile
@@ -1,0 +1,14 @@
+# Use the official Python image
+FROM python:3.9-slim
+
+# Set environment variables
+ENV PYTHONUNBUFFERED=1
+
+# Set working directory
+WORKDIR /app
+
+# Copy source code
+COPY main.py .
+
+# Run the Python script when the container starts
+CMD ["python", "main.py"]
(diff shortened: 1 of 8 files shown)


cloud_functions/test/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# Use the official Python image
+FROM python:3.9-slim
```
wlytle, line 2: If it doesn't make gdal unworkable, can we use a newer version of python? Preferably 3.13, but anything later than 3.9 is good since it goes end of life in the Fall.

infrastructure/base/modules/test_env/main.tf
```diff
@@ -0,0 +1,33 @@
...
+    TEST_ENV_VAR      = true
+  }
+}
+
+
+resource "google_vpc_access_connector" "connector" {
+  name          = "test-function-connector"
+  region        = var.gcp_region
+  network       = var.network_name
+  ip_cidr_range = "10.8.0.0/28"
```
wlytle, line 13: For my own knowledge, what's this?

infrastructure/base/modules/test_env/main.tf
```diff
@@ -0,0 +1,33 @@
...
+}
+
+
```
wlytle, line 16: Maybe, we want a module here for the data download bucket. I'm not convinced that needs to be managed in terraform, but other buckets are. 🤷

cloud_functions/test/Dockerfile
sstill88: Ya I'll definitely update this for real code (along with lots of dependencies), but I figured it was ok to print "hello world" with so I just copied a sample basic Dockerfile

infrastructure/base/modules/test_env/main.tf
sstill88: No clue lol, but you need one for a vpc connector, and I think this is a lightweight one. It includes an IP address and whatever CIDR is, and I think it's needed for communication between the different parts of a stack, but obviously this is nebulous to me.

infrastructure/base/modules/test_env/main.tf
sstill88: Yep, I'll add all add one later. Or I may just use a bucket that is created by the main stack if that exists. I just want to get something very very basic to build and then I'll add buckets and schedulers and all that good stuff.

PR #356, opened 2025-05-05
TECH-2792: Fix map layer persistence when changing languages

wlytle:
### Overview

This PR is meant to address a bug where the map layers, layer toggle switches, and map legend each do not persist state correctly or the consistently when the app language is changed. This is because map layers were being persisted by a query parameter represented by a database ID. Each version of a map layer for each language is it's own entry in the database which resulted in layers being returned for the wrong language, or - in the case of layers filtered by `locale` - no data being returned at all. In order to fix this, map layers are now being represented with a `slug` that is consistent for a single layer across all `locales`. For example the map layers for `EEZ`s has the slug `eez`  for the English, Spanish, and French version. This matches how we manage other cross language database fields, like the data sources, and popover content.

To implement this a new field of `slug` was added to the `layers` table in the database and permissions were added to allow authenticated update of layers via API calls, this allows for a script to be run to give all existing layers a slug that is unique for the layer across all languages. 

Additionally there was an error where `next-querystate` and `searchParams` were not communicating with each other. A lot of this seems to have been resolved by updating `next-querystate` (now called `nuqs`) to the latest version. This required adding some schema validation and typing for the expected query params.

### Testing instructions

Currently deployed to dev: https://30x30-dev.skytruth.org/ 
Can be tested by changing map layers, clicking the map to view pop ups, changing languages at all points in the process and making sure nothing crashes and the behavior is reasonably expected. 

### Feature relevant tickets

https://skytruth.atlassian.net/browse/TECH-2792

---


frontend/src/components/header/dropdowns/language-selector/index.tsx
```diff
@@ -9,12 +9,11 @@ const LanguageSelector: FCWithMessages = () => {
...
   const locale = useLocale();
 
-  const { pathname, asPath, query, push } = useRouter();
-
+  const { push, asPath } = useRouter();
   return (
     <Select
       value={locale}
-      onValueChange={(newLocale) => push({ pathname, query }, asPath, { locale: newLocale })}
+      onValueChange={(newLocale) => push(asPath, undefined, { locale: newLocale })}
```
wlytle, line 16: There is some bug between `next-querystate` and Next's `searchParams` that make the query params state not fully persist so [this method](https://github.com/SkyTruth/skytruth-30x30/blob/main/frontend/src/containers/map/content/map/sync-settings.ts#L37-L63) is used to preserve map settings, but there was still some weird collisions when navigating to and from non-map pages. This seems to allow for changing the language and preserving query params. Totally open to thoughts on this though. It's a bit of a guess and check solution.

frontend/src/containers/map/content/map/layers-toolbox/legend/index.tsx
```diff
@@ -72,33 +72,33 @@ const Legend: FCWithMessages = () => {
...
             .sort((a, b) => {
-              const indexA = activeLayers.indexOf(a.id);
-              const indexB = activeLayers.indexOf(b.id);
+              const indexA = activeLayers.indexOf(a.attributes.slug);
+              const indexB = activeLayers.indexOf(b.attributes.slug);
               return indexA - indexB;
             }),
         placeholderData: { data: [] },
-        queryKey: ['layers', activeLayers],
+        queryKey: ['layers', locale, activeLayers],
```
wlytle, line 82: the map legend was being cached without regard to locale, leading to some strange language behavior

frontend/src/containers/map/sync-settings.ts
```diff
@@ -1,14 +1,23 @@
-import { useQueryState } from 'next-usequerystate';
-import { parseAsJson } from 'next-usequerystate/parsers';
+import { useQueryState, parseAsJson } from 'nuqs';
```
wlytle, line 1: This is all related to the preservation of the query params when changing languages. The use of zod was to fix some typing errors that I honestly can't recall what they were now... 😬

frontend/src/components/header/dropdowns/language-selector/index.tsx
aemonm: I've had a couple of battles with the NextJS Router,  but it looks like you're on a previous version of Next and nuqs so my experience might not be useful. 

One idea - and you may have tried this already - but you could  try checking `router.isReady === true` before trying to read the Next `searchParams` to make sure you're not accessing it before the any updates have completed.

frontend/src/containers/map/content/map/layer-manager/index.tsx
```diff
@@ -19,18 +19,25 @@ const LayerManager = ({}: { cursor: CustomMapProps['cursor'] }) => {
...
+    (slug: string) => ({
+      ...(layersSettings[slug] ?? { opacity: 1, visibility: true, expand: true }),
       zoom,
     }),
     [layersSettings, zoom]
   );
 
   const layerManagerItems = useMemo(
     () =>
-      activeLayers.map((l, i) => {
```
aemonm, line 31: glad to see I'm not the only one who's one character variable habit needs breaking. 😅

frontend/src/containers/map/content/map/layer-manager/item.tsx
```diff
@@ -50,34 +63,33 @@ const LayerManagerItem = ({ id, beforeId, settings }: LayerManagerItemProps) =>
...
-
-      const { interaction_config } = data.data.attributes as LayerTyped;
+      const { interaction_config } = layer as LayerTyped;
 
       if (interaction_config?.enabled) {
         const ids = styles.map((l) => l.id);
 
-        setLayersInteractive((prev) => Array.from(new Set([...prev, id])));
+        setLayersInteractive((prev) => {
+          console.log('New ids: ', Array.from(new Set([...prev, slug])));
```
aemonm, line 72: left in a `console.log` possibly?

aemonm (approved):
lgtm

frontend/src/containers/map/content/map/layer-manager/item.tsx
wlytle: oops, good catch

frontend/src/components/header/dropdowns/language-selector/index.tsx
wlytle: Oh it actually looks the upgrade to nuqs fixed the issue here. I reverted this logic to what it used to be... There's quite a few gotchas in the router <-> nuqs world it seems 😅

PR #360, opened 2025-05-12
TECH-2513: Update landing and about page

wlytle:
### Overview

This PR primarily updates the copy for the static Landing and About pages. to facilitate this, the `SubSection` component is abstracted out of the `TwoColumnSubSection` component to allow for single column subsections.

Some images were removed as they are no longer needed. They have been stored in the 30x30 Google Drive -> Archived ->[ Deprecated Images](https://drive.google.com/drive/folders/1gJre1rbcxmwqgH4Ye0UM7nPheDJBpU3m) folder in case they are needed later.

### Designs

https://docs.google.com/document/d/1PaquS-Dq5ddwuLGADbYD5xkEAVW8h6rglr39kT4Wd7I/edit?tab=t.0

### Testing instructions

Deployed on Dev

### Feature relevant tickets

https://skytruth.atlassian.net/browse/TECH-2513


frontend/src/components/static-pages/two-col-subsection/index.tsx
```diff
@@ -54,18 +32,19 @@ const TwoColSubSection: React.FC<TwoColSubSection> = ({
         {isNumbered && (
           <span className="mb-2 font-mono text-xl md:mb-6">
             <span className="text-black">{minTwoDigits(itemNum)}</span>
-            <span className="opacity-20">-{minTwoDigits(itemTotal)}</span>
+            <span className="opacity-60">-{minTwoDigits(itemTotal)}</span>
```
wlytle, line 35: Updated to meet accessibility contrast requirements

frontend/src/components/static-pages/sub-section/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+    className={cn('flex max-h-[280px] w-full md:max-h-full', {
+      'md:mt-16': isNumbered,
+      [className]: true,
+    })}
+  >
+    {children}
+  </div>
+);
+
+SubSection.displayName = 'SubSection';
```
aemonm, line 51: What does setting the displayName like this achieve ?

frontend/src/pages/index.tsx
```diff
@@ -143,102 +148,101 @@ const Home: FCWithMessages = ({
...
+                      >
+                        {chunks}
+                      </a>
+                    ),
+                    a2: (chunks) => (
+                      <a
+                        className="underline"
+                        href={indicators?.biodiversityTextLand?.source}
+                        target="_blank"
+                        rel="noopener noreferrer"
```
aemonm, line 191: Is it still necessary to set these? I recall there being an issue with unintentionally leaking users browsing history but I thought it had been fixed in the browsers. Totally could be wrong and it might still be good for legacy browsers.

aemonm (approved):

frontend/src/components/static-pages/sub-section/index.tsx
wlytle: It's just used in dev when React throws an error to say `SubSection` broke. Usually react just gets the name of the function that defines the component, but if not it'll use `Component` if nothing is set. I'm not really sure why it fails to get the name sometimes, but I've seen it a bit lately 🤷 . The displayName is removed from the prod build though, so it shouldn't clog anything up.

frontend/src/pages/index.tsx
wlytle: Ah good point, classic copy pasta. I think Browsers have been automatically handling this for a few years. I'll yank it to clean things up.

frontend/src/components/static-pages/sub-section/index.tsx
aemonm: oh interesting - I think I've seen that happen myself now that I think about it. Thanks for the tip!

PR #361, opened 2025-05-14
Feature/data download

sstill88:
# Download raw 30x30 data

### Overview

Download and save (plus duplicate archive copies to hold onto snapshots in case that is ever useful) of each dataset needed for 30x30. The data generated by this can be found [here](https://console.cloud.google.com/storage/browser/x30-dev-data-processing;tab=objects?forceOnBucketsSortingFiltering=true&inv=1&invt=AbxUzA&project=x30-399415&prefix=&forceOnObjectsSortingFiltering=false)

### Testing instructions

- Does the code seem reasonable?
- How is the data file structure?

### Feature relevant tickets

- [TECH-2853](https://skytruth.atlassian.net/browse/TECH-2853)
- [TECH-2854](https://skytruth.atlassian.net/browse/TECH-2854)


cloud_functions/data_processing/main.py
```diff
@@ -0,0 +1,343 @@
...
+    url=PROTECTED_SEAS_URL,
+    bucket=BUCKET,
+    filename=PROTECTED_SEAS_FILE_NAME,
+    archive_filename=ARCHIVE_PROTECTED_SEAS_FILE_NAME,
+    project=PROJECT,
+    verbose=True,
+):
+    """
+    Download Protected Seas data from website and saves zipfile to a current
+    file for calculations as well as an archive file
```
wlytle, line 123: nit:
```suggestion
    Downloads Protected Seas data from website and saves zipfile to a current
    file for calculations as well as an archive file
```

cloud_functions/data_processing/requirements.txt
```diff
@@ -0,0 +1,9 @@
+pandas
+requests
+fiona
+google-cloud-storage
+tqdm
+geopandas
+fsspec
+google-api-core
+flask
```
wlytle, line 9: We should do some kind of version constraints on these, but that can happen when I move it all to Poetry

cloud_functions/data_processing/main.py
```diff
@@ -0,0 +1,343 @@
...
+
+def main(request):
+    data = request.get_json(silent=True) or {}
+    method = data.get("METHOD", "default")
+
+    if method == "dry_run":
+        print("Dry Run Complete!")
+    elif method == "download_eezs":
+        # Run in CLI via:
+        # gcloud functions call x30-dev-data --data '{"METHOD": "download_eezs"}' --region us-east1
```
wlytle, line 282: Maybe we can move these comments to a README

cloud_functions/data_processing/main.py
```diff
@@ -0,0 +1,343 @@
...
+    bucket=BUCKET,
+    filename=MPATLAS_FILE_NAME,
+    archive_filename=ARCHIVE_MPATLAS_FILE_NAME,
+    verbose=True,
+):
+    if verbose:
+        print(f"downloading MPAtlas Zone Assessment from {url}")
+
+    r = requests.get(url)
+    r.raise_for_status()
```
wlytle, line 102: All of these methods (or somewhere in the call stack)  should probably have try except statements so we can catch, log, and appropriately handle exceptions, particularly since we're making network calls and raise exceptions internally.

cloud_functions/data_processing/main.py
```diff
@@ -0,0 +1,343 @@
...
+        data = response.json()
+
+        results = data.get(val, [])
+        if not results:
+            if verbose:
+                print("All pages fetched.")
+            break
+
+        all_areas.extend(results)
+        page += 1
```
wlytle, line 183: Slight preference for a more explicit `while` condition for readability:
```suggestion
  
  def fetch_data(params):
        response = requests.get(f"{url}/{val}", params=params)
        response.raise_for_status()
        data = response.json()
        return data.get(val, [])

   results = fetch_data(params)
    while results:
        if verbose:
            print(f"Fetching page {page}...")
        all_areas.extend(results)
        page += 1
        params = {"token": pp_api_key, "per_page": per_page, "page": page}
        results = fetch_data(params)
        
     if verbose:
       print("All pages fetched.")
```

cloud_functions/data_processing/main.py
```diff
@@ -0,0 +1,343 @@
...
+
+        time.sleep(0.2)
```
wlytle, line 185: What's this for?

cloud_functions/data_processing/main.py
```diff
@@ -0,0 +1,343 @@
...
+        seamounts_url,
+        bucket,
+        seamounts_file_name,
+        archive_seamounts_file_name,
+        chunk_size=chunk_size,
+        verbose=verbose,
+    )
+
+
+def main(request):
```
wlytle, line 274: Maybe the above could live in their own file called `methods.py` or something and just main could live here. Just a thought to keep file sizes a bit smaller for readability

wlytle:
I think it's pretty standard to put all the code stuff - not DOCKERFILE - under an `src` directory. At least that's how the other cloud function is set up

sstill88:
Following what Vizzuality did for analysis, src/ would contain the referenced functions what I have in the utils dir (as well as a methods.py if we update the code to that format). The main.py, requirements.txt, and Dockerfile are all in the function's base folder. Does that sound right? And would you suggest renaming utils to src and moving params.py to src?

wlytle:
> Following what Vizzuality did for analysis, src/ would contain the referenced functions what I have in the utils dir (as well as a methods.py if we update the code to that format). The main.py, requirements.txt, and Dockerfile are all in the function's base folder. Does that sound right? And would you suggest renaming utils to src and moving params.py to src?

No that sounds good. The root has `main.py`, `Dockerfile`, etc and then `src` cna have a `utils` directory and any other helper code we need.

sstill88:
> No that sounds good. The root has `main.py`, `Dockerfile`, etc and then `src` cna have a `utils` directory and any other helper code we need.

Just trying to visualize this - so in root we have main, Dockerfile, requirements, src/ and then src/ contains utils/ (which contains gcp.py and other utils) and params.py?

wlytle (approved):
LGTM!

PR #369, opened 2025-07-03
TECH-3067: Enable source specific total_area

wlytle:
### Overview

Some of our data sources use different versions of GADM and EEZ data and, therefore, have different values for the total land and marine area of a country. This PR adds a `total_area` column to all of the statics tables (except habitat stats because that already existed as a calculated value). Additionally it adds some logic to allow our API upserting methods to pass the new field through and adds support to the front end to preferentially favor source specific total area (as opposed to what our database says the location area is).

Note: These columns will remain empty until we update and run our data pipelines with updates to pass this value through. For this reason the front end maintains fallback logic to use the current internal value of location total area.

### Feature relevant tickets

https://skytruth.atlassian.net/browse/TECH-3067
---

### DIFF
--- a/cms/config/sync/admin-role.strapi-author.json
+++ b/cms/config/sync/admin-role.strapi-author.json
@@ -1547,7 +1547,8 @@
           "pas",
           "oecms",
           "is_last_year",
-          "global_contribution"
+          "global_contribution",
+          "total_area"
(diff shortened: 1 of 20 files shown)


aemonm (approved):
LGTM!

PR #370, opened 2025-07-08
  TECH-3058: Feature Flags

wlytle:
### Overview

This PR adds a feature-flags table to the database to allow for API or Frontend programmatic feature gating or adjusting depending on records from this table.  
Key features:
* Feature flags have a generic json field called `payload` to allow for flexibility of data  delivered from the flag
* Feature flags are not returned if marked as `archived`
* Feature flags are returned based on based on date, if the optional `active_on` date is field is set
  * If the current date-time (UTC) is after `active_on` the feature will be returned
  * If there is a query parameter of `run-as-of` with a date of the format `2026-07-12T21:32:42.532Z` that date will be taken as the current date over the actual date
  * If there is a request header of `run-as-of` with a date of the format `2026-07-12T21:32:42.532Z` that date is taken as the current date over a the actual date or a `run-as-of` query parameter

Less technical documentation: https://skytruth.atlassian.net/wiki/spaces/TECH/pages/629571586/Feature+Flags

### Feature relevant tickets

[_TECH-3058_](https://skytruth.atlassian.net/browse/TECH-3058)

---

### DIFF
--- a/frontend/src/hooks/use-feature-flag.ts
+++ b/frontend/src/hooks/use-feature-flag.ts
@@ -35,7 +35,7 @@ export const useFeatureFlag = (flag: string): FeatureFlag['payload'] => {
         },
         queryKey: ['featureFlag', flag, runAsOf],
         staleTime: 'static',
-        refetchOnMount: false
+        refetchOnMount: false,
       },
     }
   );


cms/config/sync/admin-role.strapi-super-admin.json
wlytle, line 1: autogenerated config file to give full access to the feature-flags table for `super-admin` users, which is what is needed to log in to the CMS. This lets us tinker with the table and create data in the CMS

cms/config/sync/core-store.plugin_content_manager_configuration_content_types##api##feature-flag.feature-flag.json
wlytle, line 1: Autogenerated file the CMS uses to configure the table view

cms/config/sync/user-role.public.json
wlytle, line 1: Auto generated to give read access to the api for unauthenticated requests

cms/src/api/feature-flag/content-types/feature-flag/schema.json
wlytle, line 1: Autogenerated file defining the schema of the new feature-flags table  that was created via the Strapi CMS

cms/src/api/feature-flag/controllers/feature-flag.ts
```diff
@@ -0,0 +1,27 @@
+/**
+ * feature-flag controller
+ */
+
+import { factories } from '@strapi/strapi'
+
+export default factories.createCoreController('api::feature-flag.feature-flag',
+  ({ strapi }) => ({
+    async find(ctx) {
```
wlytle, line 9: This is where Strapi routes API requests from `GET /feature-flag`

cms/src/api/feature-flag/documentation/1.0.0/feature-flag.json
wlytle, line 1: Autogenerated doc file used in swagger and used by Orval in the front end to generate types.

cms/types/generated/contentTypes.d.ts
wlytle, line 1: Autogenerated types

frontend/src/types/generated/strapi.schemas.ts
wlytle, line 1: Autogenerated types, a bunch of stuff got rearranged in the file and I'm not really sure why 🤷

aemonm:
I'm having a bit of trouble with the big picture of how this works and what it's solving. (but I also don't recall how the backend works too good). 

If someone wanted to add a new datasource for display on the map, could you use a feature flag to toggle it on and off?  Could you use this to A/B test a new redesign? Or does work with more textual data you control in the CMS?

frontend/src/hooks/use-feature-flag.ts
```diff
@@ -0,0 +1,43 @@
...
+    {
+      query: {
+        select: (data) => {
+          if (data[0]) {
+            return data[0]?.attributes?.payload;
+          }
+          return null;
+        },
+        queryKey: ['featureFlag', flag, runAsOf],
+        refetchOnWindowFocus: false,
```
aemonm, line 37: oh is this enabled by default?

wlytle:
> I'm having a bit of trouble with the big picture of how this works and what it's solving. (but I also don't recall how the backend works too good).
> 
> If someone wanted to add a new datasource for display on the map, could you use a feature flag to toggle it on and off? Could you use this to A/B test a new redesign? Or does work with more textual data you control in the CMS?

The how part of this is actually much simpler than the code suggests. Just a table in the data base that returns a small bit of data conditionally. This is useful because the database/API system has a built in UI so those bits of data can be toggled on and off or updated without code changes.

As for how it's used, you pretty much nailed it. Toggling UI features, A/B testing, and it can be used for more textual data, but the idea is that it would be a limited use case. Feature Flags are in general meant to be temporary and used for the above cases to improve stability, test things out, or grant conditional access. My guess is that we'll use it primarily for feature toggling and to decouple code deploys from feature releases.

For example you could have something like
```js
return (
{ featureFlag ? <SweetNewFeature /> : <OldStandardFeature />}
);
```
And then once we're happy the feature is stable and released the above logic can be removed so we just show the cool new thing.

Or for the textual example, when we have accounts and authorizations something like
```js
return {featureFlag.roles.includes(user.role) ? <SweetNewFeature /> : <OldStandardFeature />}
```
Where the feature flag could have a payload like `{roles: ['admin', test']}` and then when we want to launch the feature we can just use the strapi API to add some the roles we want to release the feature too. 

Let me know if that doesn't make sense I'm totally happy to think about this and try to make it be as useful as possible

PR #373, opened 2025-07-31
Feature/data processing

sstill88:
### Overview

Sorry about the huge PR 😬 

1. This does a bit of an overhaul of the data_processing folder so that now there are 3 subfolders: core (parameters and utils that are specific to this analysis), methods (the actual functions run by main.py), and utils (broader utils that aren't specific to this project).
2. Adds the data processing steps - the processing steps are split up into:
- static processes (these happen infrequently, if ever, and we will not be creating cron jobs for those)
- download and process (downloads data from sources, saves raw data, and processes in any way needed for further calculations; most of these already existed, but there are a few updates, such as saving simplified geometries for quicker processing)
- generate tables (these are the processing steps that create the data to be ingested by the database
- There are two habitats (marine and terrestrial) processing steps because those each had enough steps to warrant their own files.

### Testing instructions

- Probably best to check that the numbers look right in the July version of the csvs [here](https://console.cloud.google.com/storage/browser/x30-dev-data-processing/tables?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&inv=1&invt=Ab3-5g&project=x30-399415)
- These are generated by running all of the steps in main.py that are prefixed by `generate_` (except `generate_protected_areas_table`, which is not complete yet)

### Feature relevant tickets

- [TECH-3070](https://skytruth.atlassian.net/browse/TECH-3070)
- [TECH-3084](https://skytruth.atlassian.net/browse/TECH-3084)

### DIFF
--- a/cloud_functions/data_processing/src/methods/marine_habitats.py
+++ b/cloud_functions/data_processing/src/methods/marine_habitats.py
@@ -299,7 +299,7 @@ def dissolve_multipolygons(gdf: gpd.GeoDataFrame, key: str = "WDPAID") -> gpd.Ge
     return result
 
 
-def create_marine_habitat_subtable(
+def process_marine_habitats(
     combined_regions,
     gadm_eez_union_file_name: str = GADM_EEZ_UNION_FILE_NAME,
     habitats_zipfile_name: str = HABITATS_ZIP_FILE_NAME,
--- a/cloud_functions/data_processing/src/methods/generate_tables.py
+++ b/cloud_functions/data_processing/src/methods/generate_tables.py
@@ -59,9 +59,9 @@ from src.core.processors import (
     update_mpatlas_asterisk,
 )
 
-from src.methods.marine_habitats import create_marine_habitat_subtable
+from src.methods.marine_habitats import process_marine_habitats
 
-from src.methods.terrestrial_habitats import create_terrestrial_habitats_subtable
+from src.methods.terrestrial_habitats import process_terrestrial_habitats
 
 from src.utils.gcp import (
     load_gdb_layer_from_gcs,
@@ -231,7 +231,7 @@ def generate_habitat_protection_table(
         print("loading regions")
     combined_regions, _ = load_regions()
 
-    marine_habitats = create_marine_habitat_subtable(
+    marine_habitats = process_marine_habitats(
         combined_regions,
         gadm_eez_union_file_name=gadm_eez_union_file_name,
         habitats_zipfile_name=habitats_zipfile_name,
@@ -245,7 +245,7 @@ def generate_habitat_protection_table(
         verbose=verbose,
     )
 
-    terrestrial_habitats = create_terrestrial_habitats_subtable(
+    terrestrial_habitats = process_terrestrial_habitats(
         combined_regions,
         pa_stats_filename=pa_stats_filename,
         country_stats_filename=country_stats_filename,
--- a/cloud_functions/data_processing/src/methods/terrestrial_habitats.py
+++ b/cloud_functions/data_processing/src/methods/terrestrial_habitats.py
@@ -231,7 +231,7 @@ def generate_terrestrial_biome_stats_pa(
     return pa_stats
 
 
-def create_terrestrial_habitats_subtable(
+def process_terrestrial_habitats(
     combined_regions,
     pa_stats_filename: str = PA_TERRESTRIAL_HABITATS_FILE_NAME,
     country_stats_filename: str = COUNTRY_TERRESTRIAL_HABITATS_FILE_NAME,


cloud_functions/data_processing/src/methods/download_and_process.py
```diff
@@ -347,61 +321,68 @@ def download_protected_planet(
...
-    """
-    Downloads habitat-related datasets (habitats and seamounts) and uploads them to GCS
-    as both current and archived versions.
+):
+    def create_buffer(df: gpd.GeoDataFrame) -> gpd.GeoDataFrame:
+        def calculate_radius(rep_area: float) -> float:
+            return ((rep_area * 1e6) / np.pi) ** 0.5
+
+        df = df[df["REP_AREA"] > 0].copy()
+        df["geometry"] = df.to_crs("ESRI:54009").apply(
```
wlytle, line 337: Is this something to be worried about with the weird behavior crossing 0º?

cloud_functions/data_processing/src/methods/marine_habitats.py
```diff
@@ -0,0 +1,363 @@
...
+
+        protected_area = min(df_pa_group["AREA2D"].sum(), total_area)
+
+        return {
+            "location": loc,
+            "habitat": "seamounts",
+            "environment": "marine",
+            "protected_area": protected_area,
+            "total_area": total_area,
+            # "percent_protected": 100 * protected_area / total_area if total_area > 0 else np.nan,
```
wlytle, line 69: Huh, I wonder if we should make a column for this? All the other stats tables pre-compute it, no real reason for the browser to do this math everytime 🤷

cloud_functions/data_processing/src/core/processors.py
```diff
@@ -0,0 +1,203 @@
...
+    return df
+
+
+def calculate_area(
+    df: gpd.GeoDataFrame, output_area_column="area_km2", round: None | int = 2
+) -> gpd.GeoDataFrame:
+    col = df.geometry.to_crs("ESRI:53009").area / 1e6  # convert to km2
+    if round:
+        col = col.round(round)
+    return df.assign(**{output_area_column: col})
```
wlytle, line 74: Where did we land on this method for area calculation with the whole 0º crossover chaos thing?  Or does it make sense to have two methods and use them intentionally?

cloud_functions/data_processing/src/methods/marine_habitats.py
```diff
@@ -0,0 +1,363 @@
...
+        }
+
+    def eez_location(row):
+        return row["ISO_TER1"] if isinstance(row["ISO_TER1"], str) else row["ISO_SOV1"]
```
wlytle, line 73: Is this duplicated logic in commons.py? https://github.com/SkyTruth/skytruth-30x30/pull/373/files#diff-3687eaac25a6fb53a6f1f3c9fc1b04a9661b05a7a25c2bd63ec0e6325b6589c3R40-R48

cloud_functions/data_processing/src/methods/marine_habitats.py
```diff
@@ -0,0 +1,363 @@
...
+        print("loading seamounts")
+
+    seamounts = load_zipped_shapefile_from_gcs(
+        seamounts_zipfile_name, bucket, internal_shapefile_path=seamounts_shapefile_name
+    )
+
+    if verbose:
+        print("loading eezs")
+    eez = load_marine_regions(eez_params, bucket)
+    # eez = adjust_eez_sovereign(eez, parent_country)
```
wlytle, line 85: nit: delete, if unneeded

cloud_functions/data_processing/src/methods/marine_habitats.py
```diff
@@ -0,0 +1,363 @@
...
+    multiples = gdf[gdf[key].isin(counts[counts > 1].index)]
+
+    dissolved = multiples.dissolve(by=key)
+    dissolved = dissolved.reset_index()
+    result = pd.concat([singles, dissolved], ignore_index=True)
+
+    return result
+
+
+def create_marine_habitat_subtable(
```
wlytle, line 303: I got quite confused with the naming here particularly between `create_marine_habitat_subtable` and `create_ocean_habitat_subtable` I think I see what's going on now, but maybe we can do something with the names, comments or doc stings to make this a wee bit clearer.

wlytle (approved):

cloud_functions/data_processing/src/methods/download_and_process.py
sstill88: I did a bunch of testing on this, and ended up fixing an error, but not rethinking the meridian crossing behavior. I think there are always going to be issues crossing the meridian, but with the relatively small areas, this generally works fine. Out of the 6500 buffered point PAs, there is one that returns an invalid polygon and one other that returns a re-calculated area that is more than 1% different than the REP_AREA so my thought is that this works well enough for this purpose, unless you have other thoughts.

cloud_functions/data_processing/src/methods/marine_habitats.py
sstill88: Please let me know if you'd like me to add this back in. I figured we might want it, but I commented it out so that it wouldn't mess with the DB ingest

cloud_functions/data_processing/src/methods/download_and_process.py
wlytle: If it works than that sounds fine!  My only experience with the 0 meridian bug was with an EEZ and it got it way wrong, but for tiny things if it's in the margins, that works for me

cloud_functions/data_processing/src/methods/marine_habitats.py
sstill88: We don't end up using the logic that you linked to in commons, and I'm removing it because I plan to use the geometries you are generating anyways. I also will leave this logic for now as it will also be updated for the same reason, as long as you're ok with that.

cloud_functions/data_processing/src/methods/marine_habitats.py
sstill88: done

cloud_functions/data_processing/src/utils/geo.py
```diff
@@ -0,0 +1,102 @@
...
+    else:
+        return geom
+
+
+def get_area_km2(poly):
+    wgs84 = pyproj.CRS("EPSG:4326")
+    projected_crs = pyproj.CRS("EPSG:6933")
+    transformer = pyproj.Transformer.from_crs(wgs84, projected_crs, always_xy=True)
+    projected_polygon = transform(transformer.transform, poly)
+    return projected_polygon.area / 1e6
```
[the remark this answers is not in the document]
sstill88: The difference is that `get_area_km2` calculates area for a single polygon and `calculate_area` calculates area for each polygon in a gdf. The functions in core.processors are meant to be chained in dataframe calculations (df.pipe(proc1).pipe(proc2)...) but there could be a case to be made about having `calculate_area` be based on `get_area_km2` (`calculate_area` was stolen directly from Viz)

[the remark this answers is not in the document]
sstill88: Also, I just removed `calculate_area` because it appears it isn't needed.

cloud_functions/data_processing/src/methods/marine_habitats.py
wlytle: Yup, makes perfect sense. I think I wrote that before we decided to have one EEZ file to rule them all!

cloud_functions/data_processing/src/methods/download_and_process.py
```diff
@@ -347,61 +321,68 @@ def download_protected_planet(
...
 
-    # download seamounts
-    download_and_duplicate_zipfile(
-        seamounts_url,
-        bucket,
-        seamounts_file_name,
-        archive_seamounts_file_name,
-        chunk_size=chunk_size,
-        verbose=verbose,
+    # TODO: This eliminates point PAs that have REP_AREA=0, is this what we want?
```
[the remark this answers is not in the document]
sstill88: Done!

cloud_functions/data_processing/src/methods/marine_habitats.py
wlytle: So in the new world `process_marine_habitats` does the math and whatnot to get habitats sorted out from the source data, then `create_ocean_habitat_subtable` puts the processed data into a properly formatted table, which later gets combined with the result of `create_mangroves_subtable` to eventually end up in the database? 

If I'm following that correctly, I think that's a great naming convention. Naming is so hard 😓

cloud_functions/data_processing/src/core/processors.py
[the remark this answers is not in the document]
sstill88: Added docstrings to this file

cloud_functions/data_processing/src/methods/marine_habitats.py
sstill88: In the new world, `process_marine_habitats` still encompasses `create_ocean_habitats_subtable`, but it merges all of the subtables that make up marine habitats. The subtables are:

1. ocean_habitats, which include "warmwatercorals", "coldwatercorals", "seagrasses", "saltmarshes" and come directly from a paper.
2. seamounts_subtable, which is just "seamounts"
3. mangroves_subtable, which is just "mangroves"

So process_marine_habitats creates all of those subtables and stitches them together. I had called them `create_marine_habitat_subtable` because marine_habitats is a subset of the habitats table (marine+terrestrial). But to un-confuse it with subtable (1) above, maybe I just call it `process_marine_habitats`, and just to satisfy my OCD, I'll also rename `create_terrestrial_habitats_subtable` to  `process_terrestrial_habitats`

Does that feel right?

sstill88: done!

PR #376, opened 2025-08-12
Generate  single gadm and eez files for other data processing methods

wlytle:
### Overview

This PR updates the methods to process GAMD and EEZ data such that all new definitions and mappings are included and then it saves those processed data to GCS for use by other methods.

GADM
 * Dissolves geometries on country to capture contested regions around Kashmir

EEZ
* Adds ABNJ
* Maps each location to a multipolygon of all EEZ's claimed by that country
* Also outputs a geojson of all EEZs each with multiple sovereigns listed as properties, this will be used for generating the EEZ map tiles

### Feature relevant tickets

---


cloud_functions/data_processing/src/utils/gcp.py
```diff
@@ -528,11 +536,21 @@ def read_json_df(
...
+            with tempfile.NamedTemporaryFile(suffix=".geojson", delete=False) as tmp:
+                for chunk in iter(lambda: f.read(8 * 1024 * 1024), b""):
+                    tmp.write(chunk)
+                local = tmp.name
+            try:
+                return gpd.read_file(local)
+            finally:
+                with contextlib.suppress(OSError):
+                    os.remove(local)
+
```
wlytle, line 553: Updated this because reading the terrestrial PA geojson straight into memory was crashing Docker locally, this reads geojson files to a local file as a binary stream then reads it into a dataframe before deleting the local file. Before this change reading the PA geojson took upwards of 8 Gb of CPU, after the change it was down about 1Gb

aemonm: Another trick I've found is to use line delimited geojson when possible so you can read the features line by line instead of having to load the entire file into memory in a single shot.

aemonm (commented):
a bit out of my wheelhouse, but what I understood LGTM!

cloud_functions/data_processing/src/utils/gcp.py
wlytle: This is a neat trick, thanks!

PR #378, opened 2025-08-18
Feature update for Territories

wlytle:
### Overview

This PR handles the initial deployment of code that will allow territories to appear in the UI. Because the data update for territories requires new data in the DB and new map layers, the release of the DB data is being gated behind a feature flag - `are_territories_active`  - until all data can be updated and tested for synchronization.

The bulk of this PR is setting up that gating which is done by:

### API

* Filters api calls to stats  endpoints to return country with territories  data (where applicable) so new data gaps are hidden until they can be filled, this also preserves consistency. The current set up is that USA returns dat a for USA and its territories
* Filters PA's to return data for country plus territories where appropriate

### Front End

* Generally hides territories from appearing
* Adds a section to show groups that a country belongs to in addition to just which countries belong to the group (TBD if this will stay)

### Designs

### Testing instructions

Navigate to the [dev environment ](https://30x30-dev.skytruth.org/) and as is, things should behave normally, stats should be available for sovereignties like the UK and USA.

[Add a run-as-of data ](https://skytruth.atlassian.net/wiki/spaces/TECH/pages/629571586/Feature+Flags#Schedule) for after October 1st of this year and you should see territories (maybe without stats data yet... coming in the next PR), some territories include Aruba and Puerto Rico, Sovereigns (e.g. USA* or GBR*) should be selectable and have data, while their counterparts without territories may not have data yet.

These changes don't yet apply to map interactions. 

### Feature relevant tickets

https://skytruth.atlassian.net/browse/TECH-3174

---

### DIFF
--- a/cms/src/utils/constants.ts
+++ b/cms/src/utils/constants.ts
@@ -1,5 +1,5 @@
 // TODO TECH-3174: Clean up
-export const ALL_TERRITORIES: Set<string> = new Set([
+export const ALL_TERRITORIES = new Set([
   "ARG",
   "AUS",
   "CHN",


cms/src/utils/constants.ts
```diff
@@ -0,0 +1,14 @@
+// TODO TECH-3174: Clean up
+export const ALL_TERRITORIES: Set<string> = new Set([
```
aemonm, line 2: small personal style preference, but when TypeScript can infer the types automatically (think that might be the case here but not 100% sure). I prefer to leave out the explicit type declaration.

wlytle: I like that, updated 🙂

PR #387, opened 2025-09-16
TECH-3062: Clean up query params

wlytle:
### Overview

Cleans up unused query params to try to reduce query param bloat, it also persists the run-as-of param if it's present.
More could be done here, there are layer specific query params for opacity and visibility but shorting those to `v` and `o` would require updating a bunch of config json blobs in the layer stable in the DB (downside of the cms). So I'm choosing to leave that as extra credit if we find out that we need more clean up in the url.

### Feature relevant tickets

[TECH-3062](https://skytruth.atlassian.net/browse/TECH-3062)

### DIFF
--- a/frontend/src/containers/map/sidebar/layers-panel/layers-group/index.tsx
+++ b/frontend/src/containers/map/sidebar/layers-panel/layers-group/index.tsx
@@ -7,9 +7,7 @@ import TooltipButton from '@/components/tooltip-button';
 import { Collapsible, CollapsibleContent, CollapsibleTrigger } from '@/components/ui/collapsible';
 import { Label } from '@/components/ui/label';
 import { Switch } from '@/components/ui/switch';
-import {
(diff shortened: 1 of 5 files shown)


frontend/src/containers/map/sidebar/layers-panel/layers-group/index.tsx
```diff
@@ -75,28 +71,8 @@ const LayersGroup: FCWithMessages<LayersGroupProps> = ({
...
-      })();
-
-      setLayerSettings((prev) => ({
-        ...initialSettings,
-        ...prev,
-        [layerSlug]: {
-          ...prev[layerSlug],
-          expanded: true,
-        },
-      }));
```
wlytle, line 97: None of this appeared to be used, I think it was in support of some map legend accordion feature that either got removed or never developed. Either way removing it makes the urls much cleaner by default

PR #389, opened 2025-09-22
Implement Aggregated Stats Endpoint

wlytle:
### Overview

Primarily this PR  implements the aggregated stats endpoint in the side bar widgets. It also does the following:

* Updates aggregated stats endpoint:
  * Includes related fields name (like protection level) in response and translates them according to the `locale` query parameter
  * returns multi-year response sorted by year descending
  * Defaults total area, when missing in stats record, to the location's, environment specific, total area rather than calculating it. This better matches how we handle this edge case elsewhere
* Fixes some weird UI with the bar graph when there's only 1 year of data
* Adds a hook to get the `countries` query param from the progress tracker which will be used to define what's in the custom region.

### Testing instructions

Deployed in dev, All side bar stats should populate and behave as normal


frontend/translations/en.json
```diff
@@ -173,7 +173,6 @@
       "national-level-contribution": "National level contribution",
       "global": "Global",
       "level-of-fishing-protection": "Level of Fishing Protection",
-      "highly-protected-from-fishing": "Highly protected from fishing",
```
wlytle, line 176: Moved this copy to the database since that's how all the other data types do it and it's already set up for it

PR #391, opened 2025-09-24
Update pa download

sstill88:
### Overview

This changes the way PAs are downloaded and processed. It changes which file is being downloaded (includes OECMs this time), and instead of downloading the zipfile and having to load it later, it saves simplified versions as well as a metadata only (no geometry) version of WDPA and MPA for use in PA table (the changes to the PA table that will use these metadata are not included). Another change is that this now includes point geometries as the point buffered to match the `REP_AREA`. I have also included some memory allocation tracking.  

Note - `download_and_process_protected_planet_pas` takes around ~50 minutes to complete, so will probably need to be revisited to make sure it can complete in a CloudRun.

### Feature relevant tickets

- [TECH-3017](https://skytruth.atlassian.net/browse/TECH-3071)


cloud_functions/data_processing/src/core/commons.py
```diff
@@ -261,3 +263,50 @@ def download_mpatlas_zone(
         verbose=verbose,
     )
     duplicate_blob(bucket, archive_filename, filename, verbose=True)
+
+
+def download_file_with_progress(url: str, filename: str):
```
wlytle, line 268: nit: add verbose as a param to match other functions

cloud_functions/data_processing/src/core/commons.py
```diff
@@ -261,3 +263,50 @@ def download_mpatlas_zone(
...
+
+        # Open the local file in binary write mode and create a tqdm progress bar
+        with open(filename, "wb") as file, tqdm(
+            desc=filename, total=total_size, unit="iB", unit_scale=True, unit_divisor=1024
+        ) as progress_bar:
+            # Iterate over the content in chunks and write to the file
+            for data in response.iter_content(chunk_size=8192):
+                size = file.write(data)
+                progress_bar.update(size)  # Update the progress bar with the written size
+        print(f"Download of '{filename}' completed successfully.")
```
wlytle, line 292: nit: Put this behind the verbose flag

cloud_functions/data_processing/src/core/commons.py
```diff
@@ -261,3 +263,50 @@ def download_mpatlas_zone(
...
+        return True
+    except requests.exceptions.RequestException as e:
+        print(f"Download error: {e}")
```
wlytle, line 295: nit: Use the logger so this shows up as an error log in GCP

cloud_functions/data_processing/src/core/commons.py
```diff
@@ -261,3 +263,50 @@ def download_mpatlas_zone(
...
+
+
+def print_peak_memory_allocation(func, *args, **kwargs):
+    tracemalloc.start()
+    try:
+        out = func(*args, **kwargs)
+        _, peak = tracemalloc.get_traced_memory()
+    finally:
+        tracemalloc.stop()
+    print(f"max allocated memory: {peak/(1024**3):.3f} GB")
```
wlytle, line 311: Just as a heads up there's also a nice graph of this when it runs in the cloud function or if you run it in docker, docker desktop produces one too

cloud_functions/data_processing/src/methods/download_and_process.py
```diff
@@ -132,6 +141,131 @@ def download_protected_seas(
     duplicate_blob(bucket, archive_filename, filename, verbose=True)
 
 
+def process_protected_area_geoms(
```
wlytle, line 144: This is not something to worry about, just a stylistic musing/personal preference.  There's a ton of nested functions here and it makes it a bit hard to follow the flow of the code. I like moving stuff like this to helper files or making the parent function a class instead.

cloud_functions/data_processing/src/core/commons.py
sstill88: I didn't know about the docker desktop one - but I've had issues with the cloudfunction graphs - they tend to be lagged so it's hard to tell which process caused which memory spike, miss peaks because of sampling frequency, and if there is a timeout or other reason for failure of the VM, it doesn't seem to be accurate.

Also, this helps my local testing so I can see if certain things I try drop down that allocation.

cloud_functions/data_processing/src/core/commons.py
sstill88: Can you double check that I did the logger correctly?

cloud_functions/data_processing/src/methods/download_and_process.py
sstill88: good point. I will try to clean things up in a later version to make this less confusing.

PR #419, opened 2025-12-12
extract HKG from CHN boundary

via-zhang:
Currently, in the static/GADM_processed_{tolerance}.geojson file in Google Cloud, the geometry for CHN contains HKG, which results in an overlap of the two areas. The added line will remove the HKG geometry from the CHN geometry in the countries dataset.


wlytle (approved):
LGTM!

PR #454, opened 2026-04-08
Fix Data Pipes Retry and Protected Seas Location Summing

wlytle:
## What does this PR do?

This PR does 2 unrelated things and should be 2 PR's, but here are....

1. Fixes a bug with retrying, this is done in 2 ways
  a. Set's the task queue to have a max_tries of 1, setting it to 0 turns out to be invalid so it defaulted to 100
  b. Updates the tileset generating code to retry a few times since rapid tilset uploads occasionally runs into an aws auth permission with Mapbox's upload flow
2. Fixes a bug with Protected seas stats aggregation
  a. PS added an entry in their api response for the global ocean number, so we use that instead of calculating the global number which was an overestimate because of inadvertently double counting multi-jurisdictional areas
  b. We combine some locations which PS does not, for example Svalbard and Jan Mayen is a single location the progress tracker but 2 in PS. So the ISO for Svalbard and Jan Mayen is not present in the PS API response resulting in missing data. This adds a small PS specific map to catch and sum these locations. 

Other Misc:
There was an transient error with service account application for the cloud functions. It seemed like a race condition of when permissions were applied so I slightly changed how the invoker SA is applied to the cloud functions

Task config after apply in dev:
Screenshot 2026-04-08 at 10 51 33 AM

Task config before apply:
Screenshot 2026-04-08 at 10 52 28 AM

Updated [30x30 methods](https://docs.google.com/document/d/1elV8ar1YY-MVYT8jqFfYPyHmKjGBv2hnfRmHkS2LFag/edit?tab=t.0) to include territory aggregation in Protected Seas data

## Designs

tested locally, ran full data pipeline in dev to ensure infra updated correctly


infrastructure/base/modules/cloudfunction/main.tf
```diff
@@ -117,9 +117,10 @@ resource "google_cloudfunctions2_function" "function" {
 data "google_iam_policy" "noauth" {
   binding {
     role = "roles/run.invoker"
-    members = [
-      "allUsers",
-    ]
+    members = concat(
```
sstill88: out of curiosity, what are the additional invokers?

cloud_functions/data_processing/src/methods/generate_tables.py
```diff
@@ -584,46 +584,54 @@ def get_region_stats(
...
+    # values are the iso_ter codes from Protected Seas to sum into it.
+    # This is because we combine some small locations into a single location.
+    protected_seas_iso_map = {
+        "SHN": ["ASC", "SHN", "TDC"],  # Saint Helena, Ascension and Tristan da Cunha
+        "SJM": ["JMY", "SVB"],  # Svalbard and Jan Mayen
+        "FRA": ["NAT", "CPT"],  # Clipperton Island grouped into France national
+    }
+
+    # Map different ISO codes that protected seas uses for Croatia and Global Waters
+    protected_seas["iso_sov"] = protected_seas["iso_sov"].replace({"CRV": "HRV", "OCN": "GLOB"})
```
sstill88, line 620: Nice. I think we use `global_marine_area=361000000` (though usually defined from GLOBAL_MARINE_AREA_KM2, which is 361000000) in other places, like `generate_protection_coverage_stats_table` and `generate_marine_protection_level_stats_table`, so I'm just wondering if it's better to use a consistent number across all of our tables or use theirs to be consistent with their numbers.

infrastructure/base/modules/cloudfunction/main.tf
wlytle: It's the same  scheduler invoker service account as before, but I was getting a transient  change of adding and removing account permissions that this seems to fix. It seemed like a race condition, so it just applies the iam permissions in a different way. The old implementation was in `infrastructure/base/modules/env/main.tf`

infrastructure/base/modules/cloudrun_job/main.tf
```diff
@@ -84,7 +84,11 @@ resource "google_cloud_run_v2_job" "default" {
   # Don't replace image if one exists
   lifecycle {
     ignore_changes = [
-      template[0].template[0].containers[0].image
+      template[0].template[0].containers[0].image,
+      client,
+      client_version,
+      labels,
+      template[0].labels,
```
wlytle, line 91: These are all assigned by the github actions CI/CD deploys, ignoring them just minimizes change warnings with `terraform apply`

cloud_functions/data_processing/src/methods/generate_tables.py
wlytle: This is a great question. My thoughts are that we use the source total area for all country stats so this seems to fit into that, it feels like an attempt to more faithfully represent our sources data. I have a plan to do this for MPAtlas too and we could back this number out for WDPA `(total area covered) / (percent coverage ÷ 100)`

That feels more in line with how we're doing other stats to me. But I'd love your thoughts on this.

infrastructure/base/modules/cloudfunction/main.tf
wlytle: I actually ended up reverting this. I honestly can't recall what got me into the weird state that made me think this was the fix. But I took it out and everything went smoothly and the scheduler can still run everything in staging, so seems good. Sorry for the red herring

sstill88 (approved):
Nice!

infrastructure/base/modules/cloudfunction/main.tf
wlytle: Okay I reverted my revert 🫠 
The additional invokers are the same invoker SA as always, this policy was just always applied before the binding of the service account iam policy, which re-adds it. The net affect should be the same, but the previous implementation just showed a diff in the terraform plan each time  for the creation of the policy binding.

PR #463, opened 2026-04-30
Update mpatlas global area

sashaintheocean:
## What does this PR do?

Pulls new API value from MPAtlas, saves it to GCS, uses that value to populate updated area calculations  

added two new test files, test_download_mpatlas_global.py to test flattening data api returns and test_generate_marine_protection_level_table.py to test the new table values 

```tests/methods/test_download_mpatlas_global.py::test_get_request PASSED                                                                                                                             [100%] ```

```tests/methods/test_generate_tables/test_generate_marine_protection_level_table.py::test_total_area PASSED                                                                                          [ 33%]
tests/methods/test_generate_tables/test_generate_marine_protection_level_table.py::test_combined_area PASSED                                                                                       [ 66%]
tests/methods/test_generate_tables/test_generate_marine_protection_level_table.py::test_non_GLOB_location PASSED                                                                                   [100%]
```

## Link relevant Jira tickets
https://skytruth.atlassian.net/browse/TECH-3490

### DIFF
--- a/cloud_functions/data_processing/src/methods/download_and_process.py
+++ b/cloud_functions/data_processing/src/methods/download_and_process.py
@@ -98,24 +98,26 @@ def download_mpatlas_country(
 
 def download_mpatlas_global(
     bucket: str = BUCKET,
-    project: str = PROJECT,
     url: str = MPATLAS_GLOBAL_API_URL,
     current_filename: str = MPATLAS_GLOBAL_FILE_NAME,
     archive_filename: str = ARCHIVE_MPATLAS_GLOBAL_FILE_NAME,
+    verbose: bool = True,
 ):
     response = requests.get(url)
     response.raise_for_status()
-    data = response.json()
-
-    row = {k: v for k, v in data.items() if not isinstance(v, (dict, list))}
-    for entry in data["mpaguide_status"]["total"]:
-        key = entry["key"]
-        row[f"mpaguide_total_{key}_km2"] = entry["km2"]
-        row[f"mpaguide_total_{key}_percent"] = entry["percent"]
-
-    upload_dataframe(
-        bucket, pd.DataFrame([row]), archive_filename, project_id=project, verbose=True
+    
+    if verbose:
+        logger.info(
+            {"message": f"saving MPAtlas Global API Data to gs://{bucket}/{archive_filename}"}
+        )
+    save_file_bucket(
+        response.content,
+        response.headers.get("Content-Type"),
+        archive_filename,
+        bucket,
+        verbose=verbose,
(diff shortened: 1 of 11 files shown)


cloud_functions/data_processing/src/methods/download_and_process.py
```diff
@@ -93,6 +96,29 @@ def download_mpatlas_country(
...
+):
+    response = requests.get(url)
+    response.raise_for_status()
+    data = response.json()
+
+    row = {k: v for k, v in data.items() if not isinstance(v, (dict, list))}
+    for entry in data["mpaguide_status"]["total"]:
+        key = entry["key"]
+        row[f"mpaguide_total_{key}_km2"] = entry["km2"]
+        row[f"mpaguide_total_{key}_percent"] = entry["percent"]
```
wlytle: This seems unnecessary to me. I think, as best we can, we'd like archived raw data to be raw, unprocessed data. Would be difficult to save just the raw api response either as a csv or json blob and then just grab the values you need at processing time?

cloud_functions/data_processing/src/methods/generate_tables.py
```diff
@@ -494,6 +495,24 @@ def get_group_stats(df, loc, relations, protection_level="fully-highly-protected
...
+    global_total_area = mpatlas_global["total_km2"].iloc[0]
+    # if = implementened and fully protected
+    # ih = implemented and highly protected
+    # not considering lightly protected, minimally protected, unknown,
+    # designed and unknown, or proposed/committed
+    global_protected_area = (
+        mpatlas_global["mpaguide_total_if_km2"].iloc[0]
+        + mpatlas_global["mpaguide_total_ih_km2"].iloc[0]
+    )
+
```
wlytle, line 528: The flow of this is a bit hard to follow. `global_protected_area` only gets into the `get_group_stats` function as a closure, but it's kind of buried here. Potentially easier to read would be to load the global data here and then parse it just where you need it in the stats function.

cloud_functions/data_processing/src/methods/download_and_process.py
```diff
@@ -93,6 +96,31 @@ def download_mpatlas_country(
...
+        )
+    save_file_bucket(
+        response.content,
+        response.headers.get("Content-Type"),
+        archive_filename,
+        bucket,
+        verbose=verbose,
+    )
+
+    duplicate_blob(bucket, archive_filename, current_filename, verbose=True)
```
wlytle: Are these both going to the archive now?

wlytle (approved):
LGTM!

PR #472, opened 2026-05-12
TECH-3542: Add script to backfill new locales

wlytle:
## What does this PR do?

This PR adds a manually run script to to create new DB rows for a new locale and programmatically fill those rows with the values from the default language - `en` - It also makes and join table relations to the correct locale field. e.g. A new dataset row for habitats is created in the new locale and a new rows for each habitat layers are created in the new locale and those rows are properly linked. 

For all non translated  string content the default `en` is what we want (URL's language agnostic slugs, etc) for other fields, the translated field is connected to a Localazy webhook and the content will be backfilled after it is translated. Translation fields need to have a place holder because they are often required fields.

This PR also adds `id` and `sw` to the i18n schema

## Designs

Tested locally, will test in staging with loclazy integration

## Link relevant Jira tickets

https://skytruth.atlassian.net/browse/TECH-3542


aemonm (approved):
Not super familiar with the Strapi-hijinx but this LGTM!

PR #474, opened 2026-05-14
Tech 3539: Add Bahasa and Kiswahili

wlytle:
## What does this PR do?

This PR does the initial set up to allow Bahasa an Kiswalii. Out of scope is updating translations that come from map layers (EEZ's) or are hard coded into DB layers (Country Names). Those will be handled in follow up work. This work also includes 
moving the client side localazy integration to the 30x30 V2 project that was initiated with the upgrade to Stprai V5, but that work is mostly just changing the read and write keys and CDN url for Localazy

## Designs

## Link relevant Jira tickets
https://skytruth.atlassian.net/browse/TECH-3539
As Well as 

https://skytruth.atlassian.net/browse/TECH-3537


.github/workflows/deploy-services.yml
```diff
@@ -14,11 +14,11 @@ on:
 # STAGING PHASE
 # -----------------------------
 jobs:
-  # upload_localazy:
```
wlytle, line 17: Not sure why this got commented out 😬

aemonm (approved):

cms/config/sync/core-store.plugin_localazy_content-transfer-setup.json
wlytle, line 1: All these data_tool tables were part of the about page, which doesn't exist anymore so this prevents them form syncing and translating with localazy

PR #475, opened 2026-05-15
Use node for translating location names

wlytle:
## What does this PR do?

This PR Makes use of the node and browser built in `Intl.displayName` to translate country names, along with using localazy for non-country name translations, like regions (Polar, West Asia, Areas Beyond National Jurisdiction, etc). 

This change is being implemented for several reasons:
* It allows for simpler adding of new language support in the future. 
  * No longer, is it necessary to  migrate the locations database table to add new translated name fields and also add new name fields to map tiles
* By removing translated names from map layers it will allow for a simpler future onboarding to shared geospatial map tiles (A similar solution for EEZ's is coming)
* It will align translated place names between Cerulean and 30x30

## Designs

Locally and in Dev

## Link relevant Jira tickets
https://skytruth.atlassian.net/browse/TECH-3538


frontend/src/containers/map/sidebar/main-panel/panels/details/index.tsx
```diff
@@ -63,7 +63,7 @@ const SidebarDetails: FCWithMessages = () => {
     {
       locale,
       // @ts-ignore
-      fields: ['name', 'name_es', 'name_fr', 'name_pt', 'type', 'code', 'has_shared_marine_area'],
+      fields: ['name', 'type', 'code', 'has_shared_marine_area'],
```
aemonm, line 66: Just curious what TSC was complaining about here?

frontend/src/hooks/use-location-name.ts
```diff
@@ -0,0 +1,28 @@
...
+type LocationLike = {
+  code?: string | null;
+  type?: string | null;
+};
+
+/**
+ * Returns a function that resolves a Strapi location (or a `{code, type}` literal)
+ * to a localized display name.
+ */
+export default function useLocationName() {
```
aemonm, line 60: Nice job wrapping this in a hook. Seems like it's getting a lot of mileage.

frontend/src/lib/i18n/locationName.ts
```diff
@@ -0,0 +1,92 @@
...
+ *   1. type === 'country' with `*` suffix → resolve the base country, then wrap
+ *      with the andTerritories template.
+ *   2. type === 'country' → per-locale override map, then Intl.DisplayNames with
+ *      the alpha-2 of the code.
+ *   3. type ∈ {region, custom_region, worldwide, highseas} → Localazy key
+ *      under `<type>.<code>`.
+ *   4. Anything else (inactive, inactive_region, unknown) → fall through to the
+ *      raw code so misses are visible.
+ *
+ * `t` must be scoped to the `locations` namespace.
```
aemonm: Could this be made into a hook that would call `useTranslations('locations)`  and then there would be no need to always pass in the `t` scoped to 'locations' each time ?

aemonm: Might not be worth it considering it's not used often.

aemonm (approved):

frontend/src/containers/map/sidebar/main-panel/panels/details/index.tsx
wlytle: It's a Strapi thing. Strapi expects the `fields` query param to be `fields=string` or `fields[0]=string1&fields[1]=strings2...` but It also is built to parse an array of strings. But the type for this is autogenerated and it comes up infrequently enough I've never bothered to try to fix this... probably should someday 🤷

frontend/src/lib/i18n/locationName.ts
wlytle: This was a good idea, this logic was only imported from the hook so I just added these helper functions straight into the hook.

PR #476, opened 2026-05-26
TECH-3561: Draggable legend

via-zhang:
## What does this PR do?
Replace the up and down buttons on the legend with a draggable component that preserves current layer changing and persistence rules.

## Designs
before_after_legend

1. Users can drag and drop items on the legend, and the URL for predefined layers will update accordingly.
2. Items can also be moved without a mouse: To move a layer, first click on or tab to it (which will prompt screen reader to state instructions), press Alt + Up to move up and Alt + Down to move down. Used Alt + Up/Down arrow key based on [these instructions](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-rearrangeable/).
3. Users can drag custom layers along with predefined layers.

## Issues
1. Adding a new layer from Map Layers will automatically move custom layers to the top if they are below the predefined layers (same as current version with up/down buttons).
2. Works on Window's default screen reader (Narrator) + Google Chrome and Microsoft Edge. 
3. On Mac's default screen reader (VoiceOver) + Google Chrome, Safari, and Firefox, the voiceover does not cut away when the user moves a layer.

## Relevant Jira tickets
https://skytruth.atlassian.net/browse/TECH-3561

- [x ] Merged or rebased latest code from `main`
- [ x] Tested changes on staging before Prod deployment
- [ x] Appropriate linters or formatters run
- [ x] Documentation, logging, alerts, etc appropriately updated as needed
- [ x] [Checked Demo Calendar](https://calendar.google.com/calendar/embed?src=c_650a13af470a46193658bb5cbddf79ecdb4e43bdc838d595937de6ae490704c3%40group.calendar.google.com&ctz=America%2FPhoenix) before deploying


frontend/src/containers/map/content/map/layers-toolbox/legend/index.tsx
```diff
@@ -149,172 +160,163 @@ const Legend: FCWithMessages = () => {
...
+            ...args,
+            collisionRect: { top: y, bottom: y, left: x, right: x, width: 0, height: 0 },
+          });
+        }}
+        onDragEnd={onDragEnd}
+        modifiers={[restrictToVerticalAxis, restrictToParentElement]}
+        accessibility={{
+          screenReaderInstructions: {
+            draggable:
+              'To pick up a layer, press space or enter. Use the arrow keys to move the layer up or down. Press space or enter again to drop the layer in its new position, or press escape to cancel.',
```
wlytle: Please add this to translation text. Let me know if you have questions about how to do that.

Also, I think `To move a layer, press space or enter and use the arrow keys to move the layer up or down. Press space or enter again to drop the layer in its new position, or press escape to cancel.'

wlytle: I played with this with the screen reader turned on and when dragging to a new place it says `"Draggable item {id} was moved over droppable item {id}` The id's aren't very useful it would be much nicer if it said the layer name and, if possible, rather than `dragged over layer...` if it could say something like `Layer countries was moved {above | below} terrestrial protection type`


https://dndkit.com/extend/plugins/accessibility/

frontend/src/containers/map/content/map/layers-toolbox/legend/sortable-item.tsx
```diff
@@ -0,0 +1,119 @@
...
+        <button
+          ref={setActivatorNodeRef}
+          {...listeners}
+          type="button"
+          onKeyDown={handleKeyDown}
+          aria-label={t('drag-to-reorder-layer', {layer: title})}
+          className="mt-[3px] shrink-0 cursor-grab touch-none rounded-sm text-gray-400 focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-black focus-visible:ring-offset-1 active:cursor-grabbing"
+        >
+          <Menu className="h-4 w-4" aria-hidden />
+        </button>
```
wlytle: Nit: It conceptually feels like this button should live with the header component

frontend/src/containers/map/content/map/layers-toolbox/legend/index.tsx
```diff
@@ -149,172 +159,205 @@ const Legend: FCWithMessages = () => {
...
-
-      setAllActiveLayers((prev) =>
-        prev.toSpliced(layerIndex, 1).toSpliced(layerIndex + delta, 0, layerSlug)
-      );
+      const neighborLabel = getLayerLabel(allActiveLayers[toIndex]);
+      const direction = toIndex > fromIndex ? 'below' : 'above';
+      const reordered = arrayMove(allActiveLayers, fromIndex, toIndex);
+      setAllActiveLayers(reordered);
+      setPredefinedMapLayers(reordered.filter((layerSlug) => !customLayers[layerSlug]));
+      setAnnouncement(`${getLayerLabel(slug)} moved ${direction} ${neighborLabel}`);
```
wlytle: This needs to be constructed with a translated string. Let me know if you some help sorting that out

frontend/src/containers/map/content/map/layers-toolbox/legend/sortable-item.tsx
via-zhang: Moved to `LegendItemHeader`!

frontend/src/containers/map/content/map/layers-toolbox/legend/index.tsx
via-zhang: Set as translated string

wlytle (approved):
LGTM

PR #480, opened 2026-06-04
Fix CRC Stats

wlytle:
## What does this PR do?

This PR fixes several bugs with Climate Resilient Corals stats and some related other habitat stats bugs

- Uses the EEZ - Land union to calculate total habitat and coverage for Corals
  - Many corals were between the EEZ and country boundaries and were getting left off
- Switches to using the Marine regions EEZ-Land union file instead of our home grown version
  - Our version accidentally collapsed small EEZ's entirely within other EEZs like Columbia's Seranillo bank
- Updates the area calculation for raster Cells to match the mathematical basis for the vector area calculation (claude's doing)
- Calculates total global area of CRC's rather than rolling up from countries to avoid double counting overlapping EEZs
- Fixes a small but mighty bug that was causing habitat stats to be appended in the DB instead of overwriting stats for the same year. This caused multiple stats for 2026, and because of how those are fetched for the UI all of those values were being summed for display resulting in WAY too big of numbers
  - Old appended entries have been cleared out

Tested locally and is deployed in dev

- [ ] Merged or rebased latest code from `main`
- [ ] Appropriate linters or formatters run
- [ ] Documentation, logging, alerts, etc appropriately updated as needed


cms/src/api/habitat-stat/services/habitat-stat.ts
```diff
@@ -6,10 +6,10 @@ import { factories } from '@strapi/strapi';
 
 export default factories.createCoreService('api::habitat-stat.habitat-stat', ({ strapi }) => ({
   async getHabitatStatMap(year: number): Promise<IDMap> {
+    // NOTE: do not filter by locale here. habitat-stat is not localized
     const habitatStats = await strapi.documents('api::habitat-stat.habitat-stat').findMany({
-      filters: { 
+      filters: {
         year,
-        locale: 'en',
```
wlytle, line 12: Since habitat stats isn't localized, this caused matching stats records for the year to return null since no stat record has `locale = 'en'` and so updated records were appended instead of overwriting

cloud_functions/data_processing/src/methods/static_processes.py
```diff
@@ -380,53 +380,71 @@ def _assign_terrs_and_sovs(row):
...
     verbose: bool = True,
 ):
-    eez_file_name = eez_file_name.replace(".geojson", f"_{tolerance}.geojson")
-    gadm_file_name = gadm_file_name.replace(".geojson", f"_{tolerance}.geojson")
-    gadm_eez_union_file_name = gadm_eez_union_file_name.replace(".geojson", f"_{tolerance}.geojson")
-
-    eez = read_json_df(bucket, eez_file_name, verbose=verbose)
-    gadm = read_json_df(bucket, gadm_file_name, verbose=verbose).to_crs(eez.crs)
+    """Build the per-location land/EEZ union from the Marine Regions EEZ-land-union.
+
```
sstill88, line 392: nit: should we track what's changed in the method description? `process_eez_gadm_unions` no longer exists so probably no reason to keep note of it right?

cloud_functions/data_processing/src/utils/geo.py
```diff
@@ -6,15 +6,44 @@
...
+    return (
+        WGS84_A_M**2
+        * (1 - WGS84_E2)
+        * (sin_lat / (2 * (1 - WGS84_E2 * sin_lat**2)) + np.arctanh(e * sin_lat) / (2 * e))
+    )
 
 
 def compute_pixel_area_map_km2(transform: Affine, width: int, height: int) -> np.ndarray:
     """
     Computes a 2D array of pixel areas (in km²) for a georeferenced image in geographic CRS.
```
sstill88, line 40: Not an issue, but is there substantial enough difference between the simplified area calculation and the new one to notice it in the results?

cloud_functions/data_processing/src/methods/static_processes.py
wlytle: Yeah great point. Overly "helpful" Claude on this one. I've deleted a lot of this doc string

sstill88 (approved):
Looks good

cloud_functions/data_processing/src/utils/geo.py
wlytle: It is noticeable, but small. the biggest difference I saw was a reduction of about 2%-ish and given that we are over on all these area calculations (which I'm more convinced is because we have too course of data) I'm taking the tiny reduction in calculated area where I can get it 😅

PR #485, opened 2026-08-04
Iho water bodies

sstill88:
## Description

IHO sea areas (plus the full Mediterranean Region - 11 IHO sea areas dissolved) are added and stats are computed over each. Dissolved Mediterranean polygon looks like this:
image

## Other changes:
- move `download_protected_planet_country` to a follow up of `download_protected_planet_pas` instead of its own monthly trigger because the former triggers a method that calls `compute_iho_protection_coverage`, which depends on `WDPA_MARINE_FILE_NAME` (created by  `download_protected_planet_pas`)
- bugfix: rename GID_0 --> location in `generate_terrestrial_biome_stats_country`
- updated some file loading so it will automatically choose parquet or geojson based on extension
- renamed country --> location since we now include IHO water bodies and "country" no longer makes sense
- moved `generate_protection_coverage_stats_table` to `LONG_RUNNING_TASKS` because it now exceeds time limit

## Notes before running on prod:
- **Seed Protected Seas first.** Run `seed_protected_seas_sites` (one-time,
  from the local LFP JSON dir) BEFORE the first monthly run. `download_protected_seas`
  now calls `update_protected_seas_data`, which reads the existing
  `protected_seas/protected_seas_sites.parquet` — it will error if the seed
  file doesn't exist yet.

- **Regenerate static files (in this order — later steps depend on earlier ones):**
  - `download_iho_sea_areas` → `process_iho_sea_areas` (adds Mediterranean Region;
    writes `static/iho_sea_areas_processed_{tolerance}.parquet`)
  - `process_mangroves` — now unions mangroves over IHO regions too, so it must
    run *after* the IHO file exists; also writes a NEW blob
    `static/mangroves_by_region.geojson` (old `mangroves_by_country.geojson`
    is now orphaned)
  - regenerate the coral / marine-habitat static tables that switched
    country→location, so their schema matches the new stats tables

- **Confirm IHO regions are registered as locations.** `generate_locations_table`
  still builds only from EEZ + GADM — it does not add IHO MRGID / "MEDI" rows.
  The new stats rows are keyed by `location = MRGID` (or "MEDI").

- tested full publisher pipeline on Staging

## Link relevant Jira tickets
- [TECH-3627](https://skytruth.atlassian.net/browse/TECH-3627)

- [ ] Merged or rebased latest code from `main`
- [ ] Appropriate linters or formatters run
- [ ] Documentation, logging, alerts, etc appropriately updated as needed


cloud_functions/data_processing/src/core/commons.py
```diff
@@ -137,15 +138,30 @@ def download_and_duplicate_zipfile(
     duplicate_blob(bucket, archive_blob_name, blob_name, verbose=True)
 
 
+def add_tolerance_suffix(filename: str, tolerance) -> str:
```
wlytle, line 141: ❤️

cloud_functions/data_processing/src/methods/generate_tables.py
```diff
@@ -592,11 +432,50 @@ def get_group_stats(
...
+    # Load the current Protected Seas sites.
+    ps_sites = read_parquet_from_gcs(bucket, sites_file_name, verbose=verbose)
+
+    # Keep only the highly protected sites, then merge them into a single
+    # geometry so we can measure their overlap with each IHO sea area.
+    highly = ps_sites[ps_sites["fishing_protection_level"] == "highly"]
+
+    # Repair any invalid site geometries before the overlay/dissolve.
+    if verbose:
+        logger.info({"message": "making geometries valid"})
```
wlytle: nit:

```suggestion
        logger.info({"message": "making Protected Seas geometries valid"})
```
The logs get a bit crowded, extra defining words are helpful when things go wrong

cloud_functions/data_processing/src/methods/marine_habitats.py
```diff
@@ -159,40 +188,62 @@ def get_group_stats(df, loc, relations, global_mangrove_area):
...
-        logger.info({"message": "loading pre-processed mangroves"})
-    mangroves_by_country = read_json_df(bucket, mangroves_by_country_file_name, verbose=True).pipe(
-        clean_geometries
+        logger.info({"message": "loading IHO sea areas"})
+    iho = read_parquet_from_gcs(
+        bucket, add_tolerance_suffix(iho_file_name, tolerance), verbose=verbose
     )
+    iho["location"] = iho["MRGID"].astype(str)
+
+    regions = gpd.GeoDataFrame(
```
wlytle: nit: variable name change to `locations`

cloud_functions/data_processing/src/methods/protected_seas.py
```diff
@@ -0,0 +1,278 @@
...
+    BUCKET,
+    PROJECT,
+    PROTECTED_SEAS_SITES_FILE_NAME,
+)
+from src.utils.gcp import duplicate_blob, read_parquet_from_gcs, upload_gdf
+from src.utils.logger import Logger
+
+logger = Logger()
+
+BASE_URL = "https://map.navigatormap.org/api"
```
wlytle: Can probably move this to `params.py`

cloud_functions/data_processing/src/methods/protected_seas.py
```diff
@@ -0,0 +1,278 @@
...
+# Map Level of Fishing Protection (lfp) to our fishing_protection_level buckets.
+# lfp=0 (unprotected) has no bucket and maps to NaN.
+FISHING_PROTECTION_MAPPING = {1: "less", 2: "less", 3: "moderately", 4: "highly", 5: "highly"}
+
+
+def force_2d(geom):
+    return transform(lambda x, y, z=None: (x, y), geom)
+
+
+def seed_protected_seas_sites(
```
wlytle, line 35: I'm not quite sure what this function does and it doesn't appear to be called anywhere. Do we need it?

cloud_functions/data_processing/src/methods/protected_seas.py
```diff
@@ -0,0 +1,278 @@
...
+    logger.info(
+        {
+            "message": (
+                f"{len(changed_ids)} changed and {len(removed_ids)} removed site(s) "
+                f"since {changed_since}; fetching details for changed sites"
+            )
+        }
+    )
+
+    gdfs = []
```
wlytle, line 212: What does gdfs stand for?

cloud_functions/data_processing/src/methods/protection_coverage.py
```diff
@@ -0,0 +1,369 @@
...
+    iho_proj = iho.to_crs(epsg=6933)
+    pas_proj = pas.to_crs(epsg=6933)
+
+    iho_proj["geometry"] = iho_proj.geometry.apply(make_valid)
+    pas_proj["geometry"] = pas_proj.geometry.apply(make_valid)
+
+    sindex = pas_proj.sindex
+    results = []
+
+    for _, region in iho_proj.iterrows():
```
wlytle: nit: Maybe change region to `sea`?

cloud_functions/data_processing/src/methods/protection_coverage.py
```diff
@@ -0,0 +1,369 @@
...
+                    "coverage": 0.0,
+                    "pas": 0.0,
+                    "oecms": 0.0,
+                    "protected_areas_count": 0,
+                }
+            )
+            continue
+
+        actual = pas_proj.iloc[candidates]
+        actual = actual[actual.intersects(region.geometry)]
```
wlytle: Probably obvious, but what's going on here with candidates vs actual? Is candidates just a lower resolution check that's faster?

cloud_functions/data_processing/src/methods/publisher.py
```diff
@@ -191,10 +194,6 @@ def monthly_job_publisher(task_config, long_running_task_list=None, verbose=True
             "METHOD": "download_protected_seas",
             **task_config,
         },
-        {
-            "METHOD": "download_protected_planet_country",
-            **task_config,
-        },
```
wlytle, line 197: Why did this get removed?

cloud_functions/data_processing/src/methods/static_processes.py
```diff
@@ -434,12 +439,63 @@ def process_eez_land_union(
...
+    verbose: bool = True,
+):
+    if verbose:
+        logger.info({"message": f"loading IHO sea areas from {iho_params['zipfile_name']}"})
+
+    iho = load_marine_regions(iho_params, bucket)
+    iho = stitch_mediterannean(iho)
+    original_geometry = iho["geometry"].copy()
+
+    for tolerance in tolerances:
```
wlytle, line 486: Do we need multiple tolerances for this? Feels like maybe we can just do the one we need.

cloud_functions/data_processing/src/utils/gcp.py
```diff
@@ -188,9 +188,26 @@ def download_zip_to_gcs(
...
+            # Submit it a second time with the honeypot field left empty.
+            if "text/html" in response.headers.get("content-type", ""):
+                import re
+
+                honeypot = re.search(r'<input name="(firstname-[^"]+)"', response.text)
+                confirm_data = dict(data)
+                if honeypot:
+                    confirm_data[honeypot.group(1)] = ""
+                if verbose:
+                    print("confirmation form detected, resubmitting")
```
wlytle: switch to logger

cloud_functions/data_processing/src/utils/gcp.py
```diff
@@ -188,9 +188,26 @@ def download_zip_to_gcs(
...
-            response = requests.post(
+            session = requests.Session()
+            response = session.post(
                 url, params=params, data=data, headers=headers, allow_redirects=True
             )
+            response.raise_for_status()
+            # Some endpoints return an HTML confirmation form on the first POST.
+            # Submit it a second time with the honeypot field left empty.
+            if "text/html" in response.headers.get("content-type", ""):
+                import re
```
wlytle: Maybe move this import to the top to match the others

cloud_functions/data_processing/src/utils/gcp.py
```diff
@@ -188,9 +188,26 @@ def download_zip_to_gcs(
 
     try:
         if data is not None:
-            response = requests.post(
+            session = requests.Session()
```
wlytle, line 198: Was this all added to get marine regions to download?

cloud_functions/data_processing/src/utils/gcp.py
```diff
@@ -676,6 +702,17 @@ def read_json_from_gcs(bucket_name: str, filename: str, verbose: bool = True) ->
         return json.load(f)
 
 
+def write_json_to_gcs(bucket_name: str, filename: str, data: dict, verbose: bool = True) -> None:
```
wlytle: Is this different enough from `save_json_to_gcs` on L291 to need a new function?

cloud_functions/data_processing/src/utils/gcp.py
```diff
@@ -676,6 +702,17 @@ def read_json_from_gcs(bucket_name: str, filename: str, verbose: bool = True) ->
...
+    fs = gcsfs.GCSFileSystem()
+    gcs_path = f"gs://{bucket_name}/{filename}"
+
+    if verbose:
+        print(f"Writing JSON to {gcs_path}")
```
wlytle: nit: switch to logger

wlytle (commented):
This is already a big PR so I don't want to push it, but it might be nice to add some tests for all this new logic.

cloud_functions/data_processing/src/methods/publisher.py
```diff
@@ -390,12 +406,14 @@ def dispatch_publisher(
...
             )
-            if tolerance == TOLERANCES[0]:
+            if tolerance == terrestrial_tolerance:
                 step_list = [
                     "generate_protected_areas_table",
                     "generate_terrestrial_biome_stats",
                     "generate_eez_minus_mpa",
                 ]
+            if tolerance == marine_tolerance:
+                step_list = ["download_protected_planet_country"]
```
wlytle, line 416: I'm a bit confused by this. Can you help me understand why this got moved here and only for a single `marine_tolerance`

cloud_functions/data_processing/src/methods/publisher.py
wlytle: NVM I see where it got moved to, and added my question there

cloud_functions/data_processing/src/methods/protected_seas.py
wlytle: Wait I found it. Maybe this should be one of the static cases in the publisher? Or at least a comment here explaining what it is and why it's not called by the code.

cloud_functions/data_processing/src/methods/marine_habitats.py
sstill88: Is there a reason for this preference? I think we use `location` throughout

cloud_functions/data_processing/src/methods/protected_seas.py
sstill88: renamed to `PROTECTED_SEAS_SITES_URL` and moved to params

cloud_functions/data_processing/src/methods/protected_seas.py
sstill88: The reason I didn't add it to the publisher is that this task runs using a local file. The file is one shared with me directly from Protected Seas. I felt like anything run in the publisher shouldn't require local files. Do you think I should add it to publisher anyways?

I did add some docstrings though.

cloud_functions/data_processing/src/methods/protected_seas.py
sstill88: the plural of gdf (GeoDataFrame)

cloud_functions/data_processing/src/methods/marine_habitats.py
wlytle: Just that `regions` are a specific type of location in the context of this code and we seem to be migrating toward `location` as a replacement for `country` or `region` unless it's specifically referring to one of those.

cloud_functions/data_processing/src/methods/protection_coverage.py
sstill88: done

cloud_functions/data_processing/src/methods/protected_seas.py
wlytle: No, I guess if it has to run locally it doesn't make sense in the publisher. That should just all be explained so it's clear why this is here.

cloud_functions/data_processing/src/methods/protection_coverage.py
sstill88: We filter by bounding box, which quickly eliminates most protected areas (bbox is lightweight for filtering). We then do the more expensive exact geometry-intersection check on the much smaller candidate set. Adding some comments to make that clear.

cloud_functions/data_processing/src/methods/protected_seas.py
wlytle: hahaha, well that makes sense!

cloud_functions/data_processing/src/methods/static_processes.py
sstill88: Is it cool if I leave this for now and take this on in the next sprint with the task to remove tolerances altogether?

wlytle: Yup! This is what got me thinking about that!

cloud_functions/data_processing/src/utils/gcp.py
sstill88: Done. I also updated a bunch of other print statements to logger throughout this file.

cloud_functions/data_processing/src/utils/gcp.py
sstill88: done

cloud_functions/data_processing/src/utils/gcp.py
sstill88: yep. Not really sure what changed but it would not download. Full disclosure, most of this was added by Claude.

cloud_functions/data_processing/src/utils/gcp.py
sstill88: done

cloud_functions/data_processing/src/methods/publisher.py
sstill88: `download_protected_planet_country` requires protected planet PAs in the marine tolerance (now that we've added IHO boundaries), whereas the other ones needed terrestrial tolerance. In any case, this will be a non-issue after we remove tolerance once and for all.

cloud_functions/data_processing/src/utils/gcp.py
sstill88: I don't know why I added this - it doesn't get used anyways. Removed!

cloud_functions/data_processing/src/methods/marine_habitats.py
sstill88: OH! I am a dummy. I thought you meant change "location" to "locations", not "regions" to "locations" 🤦‍♀️ Fixing now.

cloud_functions/data_processing/src/methods/protected_seas.py
sstill88: ok, hopefully the docstrings explain well enough. I'll also update the readme.

cloud_functions/data_processing/src/utils/gcp.py
wlytle: Marine regions is (intentionally?) set up to not be used this way, so not terribly surprising

cloud_functions/data_processing/src/utils/gcp.py
wlytle: :old-man-yells-at-claude:

cloud_functions/data_processing/src/methods/protected_seas.py
wlytle: Looks great, thanks for the clarification!

wlytle (approved):
LGTM!

